PR #278, opened 2013-12-23
Option to track console progress

Figedi:
I totally forgot about pull-requesting this. Anyways I've found it handy to be able to track the output of wkhtmltopdf from the console, in order to display the overall progress of the pdf creation.

This could be used to implement realtime-updates for the enduser, thus providing a much nicer feedback. Since the actual implementation of the user-feedback is application-dependent, one can use the newly created `progress`-Proc in the pdf options hash, which gets called every time the console has a new line/update (wkhtmltopdf supports a 0-100% progress bar from command line which you can easily parse with regex or whatever you like).

Note that this is **NOT a windows** feature, as windows doesn't have PTY and I currently don't know how to implement this on windows machines. On the other hand, I doubt that this is a great loss.


matzke:
+1 

Could you add a simple example for option and using the output for progress!?

matkoniecz:
@maintainers Can you consider either accepting and merging after rebasing or rejecting it?

unixmonkey:
Yes. I'll reconsider merging if this could be brought up-to-date, and most of the added lines are put into a separate method that is only called when an option is passed to opt-in to this behavior.

Something like: `WickedPdf.options[:show_console_progress:] = true`

PR #301, opened 2014-05-09
Add option to set manually wkhtml2pdf output file

j15e:
For exemple :

```ruby
html = '<h1>test</h1>'
ouput_file = File.new(File.join(Rails.root, 'public', 'export-2013.pdf'), 'w+')
WickedPdf.new.pdf_from_string(html, output_file: output_file))
```

Might be useful in case you run this script from a web server that will get kill after ex. 10 sec. timeout. With this option set, the PDF file will get written where you want directly from the command line without using an other temporary file via Ruby (that might get kill if it is too long to generate).

I renamed the variable `output_file` as this is the option name in wkhtml2pdf documentation.

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -79,8 +79,14 @@ class WickedPdf
   rescue Exception => e
     raise "Failed to execute:\n#{command}\nError: #{e}"
   ensure
-    string_file.close! if string_file
-    output_file.close! if output_file && !return_file
+    string_file.close if string_file
+    if output_file && !return_file
+      if output_file.is_a?(Tempfile)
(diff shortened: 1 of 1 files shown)


j15e:
The difference is that you may specify the exact path (non-temporary) of the PDF, so that the `wkhtmltopdf` command will be instructed to output there directly (ex. `public/reports/report-123.pdf`). This way you do not have to copy the temporary file after the processing is completed (which in our case took a long time and the ruby process would get killed before having the time to copy the temporary to its final output). 

Obviously, the problem we were having should be dealt with background worker. But I think the option to fully manage the way the binary get executed would still be useful.

PR #346, opened 2014-10-24
Include path configuration for wkhtmltopdf-binary gem

deftOfCenter:
Just spent a good hour trying to find this solution.  Most of the options I saw gave the instruction "put in the right path".  I figured there was a way to grab the 'right path', but wasn't quite well versed to figure it out.  I almost gave up trying to figure it out and went the route of downloading the binary, then extracting it, and putting it into my /bin.  Not very future proof.

 This will automatically update the path when the gem is updated, and the script it points to works by looking at the system to determine the correct binary.  Much cleaner.

Hope it helps!

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,13 +35,16 @@ gem 'wkhtmltopdf-binary'
 
 To your Gemfile and run `bundle install`.
 
-If your wkhtmltopdf executable is not on your webserver's path, you can configure it in an initializer:
+Then configure it to look in the initializer:
 
 ```ruby
 WickedPdf.config = {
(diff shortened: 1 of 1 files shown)


unixmonkey:
I've updated the generated initializer to have `exe_path` commented out by default in [this commit](https://github.com/mileszs/wicked_pdf/commit/16b87f3442bda352df60252342492bf7676621c4), along with some notes that it generally isn't needed, and the tip on using Gem.bin_path.

I'm closing this issue for now, but if you think I'm missing something important, please reopen.

PR #379, opened 2015-01-27
WickedPdfHelper.add_extension change

macarenomarco:
Changed the add_extension method of WickedPdfHelper so it doesn't add the extension if the filename already has it.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper.rb
@@ -6,7 +6,7 @@ module WickedPdfHelper
   end
 
   def self.add_extension(filename, extension)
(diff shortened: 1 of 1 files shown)


macarenomarco:
In my rails app the main css file has an css.scss extension, which means that if I use application.css as the source name, the scss never gets referenced. Maybe a better fix would be to include an extra parameter? In that way the original behavior can still be relied upon.

unixmonkey:
Thanks for clarifying! I haven't considered that.

I try and test a few configurations, but this may be a better way to deal with that. Though I still don't love that it may have false positives, it is probably better than not being able to find the right assets, or make it so they can't precompile.

PR #394, opened 2015-03-18
Fixed check for HTTPS url instead of just HTTP url

scouttyg:
This one was a huge pain in the butt, and took a long time to find. But I found it!

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper.rb
@@ -95,7 +95,7 @@ module WickedPdfHelper
       protocol = WickedPdf.config[:default_protocol] || 'http'
       if source[0, 2] == '//'
         source = [protocol, ':', source].join
-      elsif !source[0, 7].include?('://')
+      elsif !source[0, 8].include?('://')
         source = [protocol, '://', source].join
       end
       source
--- a/test/functional/wicked_pdf_helper_assets_test.rb
+++ b/test/functional/wicked_pdf_helper_assets_test.rb
@@ -6,18 +6,23 @@ class WickedPdfHelperAssetsTest < ActionView::TestCase
   include WickedPdfHelper::Assets
 
   if Rails::VERSION::MAJOR == 4
-    test 'wicked_pdf_asset_path should return an url when assets are served by an asset server' do
+    test 'wicked_pdf_asset_path should return a url when assets are served by an asset server' do
       expects(:asset_pathname => 'http://assets.domain.com/dummy.png')
       assert_equal 'http://assets.domain.com/dummy.png', wicked_pdf_asset_path('dummy.png')
     end
 
-    test 'wicked_pdf_asset_path should return an url with a protocol when assets are served by an asset server with relative urls' do
+    test 'wicked_pdf_asset_path should return a url when assets are served by an asset server using HTTPS' do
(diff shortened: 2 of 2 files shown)


lib/wicked_pdf/wicked_pdf_helper.rb
```diff
@@ -95,7 +95,7 @@ def set_protocol(source)
       protocol = WickedPdf.config[:default_protocol] || 'http'
       if source[0, 2] == '//'
         source = [protocol, ':', source].join
-      elsif !source[0, 7].include?('://')
```
scouttyg: Imagine you have a variable named source, defined as follows:

```source = "https://cdn.example.com/assets/application_fha2215asfahi21f1.css"```

What would ```source[0,7]``` be? In this case, "https:/", which would not include "://", and so we'd get an asset like:

```"http://https://cdn.example.com/assets/application_fha2215asfahi21f1.css"```

File: lib/wicked_pdf/wicked_pdf_helper.rb [hunk 2]
```diff
@@ -95,7 +95,7 @@ def set_protocol(source)
...
+      elsif !source[0, 8].include?('://')
```
scouttyg: Lets give one more character to get that HTTPS goodness!

lib/wicked_pdf/wicked_pdf_helper.rb
scouttyg: This seems counter intuitive, since ```source[0] == "h"```, ```source[7] == "/"``, and ```source[8] == "c"```, but ```source[0,y]``` is giving us a string range, specifically the first y characters  -- ```source[0,0] ==  ""```, ```source[0,1] == "h"```, and ```source[0,7] == "https:/"```

test/functional/wicked_pdf_helper_assets_test.rb
```diff
@@ -6,18 +6,23 @@ class WickedPdfHelperAssetsTest < ActionView::TestCase
   include WickedPdfHelper::Assets
 
   if Rails::VERSION::MAJOR == 4
-    test 'wicked_pdf_asset_path should return an url when assets are served by an asset server' do
+    test 'wicked_pdf_asset_path should return a url when assets are served by an asset server' do
```
scouttyg: Just more of a semantics thing, but it should probably be "an" url (see: http://itknowledgeexchange.techtarget.com/writing-for-business/which-is-correct-a-url-or-an-url/)

test/functional/wicked_pdf_helper_assets_test.rb
```diff
@@ -6,18 +6,23 @@ class WickedPdfHelperAssetsTest < ActionView::TestCase
...
       expects(:asset_pathname => 'http://assets.domain.com/dummy.png')
       assert_equal 'http://assets.domain.com/dummy.png', wicked_pdf_asset_path('dummy.png')
     end
 
-    test 'wicked_pdf_asset_path should return an url with a protocol when assets are served by an asset server with relative urls' do
+    test 'wicked_pdf_asset_path should return a url when assets are served by an asset server using HTTPS' do
```
scouttyg: Quick test to demo out this error -- if we change it back to source[0,7] in the helper then this test will fail, changing it to source[0,8] in the helper will make it pass again :confetti_ball:

PR #418, opened 2015-06-04
Remove rails dependency

mperham:
This is a big speed bump for people who want to use wicked_pdf in a simple Ruby script.  For instance, now I need to install libxml2 because rails depends on nokogiri which depends on libxml2.

Since anyone who wants to use wicked_pdf in a Rails app will, by definition, already have rails loaded, it would seem there's no need to require it in the gemspec.  Confirm/deny?

### DIFF
--- a/wicked_pdf.gemspec
+++ b/wicked_pdf.gemspec
@@ -18,7 +18,6 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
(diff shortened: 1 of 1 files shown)


unixmonkey:
Thanks! It does seem that many people like to use pdf_from_string. It doesn't support all the rendering-specific options, but maybe that doesn't matter to those that use it that way.

I added it back as a development_dependency, though. That lets me run the tests (which generate a Rails app). That doesn't interfere with what you are trying to do, does it?

mperham:
Thanks!  It shouldn't cause a problem.

PR #436, opened 2015-08-05
Add missing `pdf =` assignment with example.

joshuapinter:
Unless I'm missing something, this should be here.

Great lib, by the way. :)


unixmonkey:
Thanks for the assist!

joshuapinter:
:+1:

PR #444, opened 2015-08-27
Uses a duplicate options hash when parsing options

### LINKED ISSUE #441: Margin options in initializers/wicked_pdf.rb do not stick after first refresh
I'm using the WickedPdf Middleware to be able to print out specific pages as PDFs. Every page will have the same PDF options so I have put them into `config/initializers/wicked_pdf.rb`. After adding these options and restarting the server, the PDF renders with the correct margins. When I refresh the same page, the custom margins are now gone and are back to the default margins. I can't get the margins to show again until I restart the server again, which is less than ideal :smile: 

In my `config/initializers/wicked_pdf.rb`:
```ruby
WickedPdf.config = {
  exe_path: '/usr/local/bin/wkhtmltopdf',
  margin: { top:    5,
            bottom: 5,
            left:   5,
            right:  5 },
  page_size: 'Tabloid',
  orientation: 'Landscape'
}
```

In my `config/application.rb`:
```ruby
require File.expand_path('../boot', __FILE__)
require 'rails/all'
Bundler.require(*Rails.groups)

module MyApp
  class Application < Rails::Application
    config.active_record.raise_in_transactional_callbacks = true
    config.eager_load_paths += ["#{config.root}/lib"]
    require 'wicked_pdf'
    config.middleware.use WickedPdf::Middleware
  end
end
```

a-maas:
This will resolve #441.

The issue was that that parts of the options hash in memory was being deleted which wicked_pdf parsed the options, specifically in my case, the [margins hash](https://github.com/mileszs/wicked_pdf/blob/be360d648388c48534263d7d3c7208ca43e58db8/lib/wicked_pdf.rb#L295). This resulted in those options going away for subsequent pdf generations after Rails initialization. This was specifically an issue when using the wicked_pdf middleware because the options are passed in once at initialization. Instead of deleting the in memory options, I'm making a duplicate of it and using that. I tested it on a local application and it resolved my issues.

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -49,6 +49,7 @@ class WickedPdf
   end
 
   def pdf_from_string(string, options = {})
+    options = options.dup
     # merge in global config options
     options.merge!(WickedPdf.config) {|key, option, config| option}
     string_file = WickedPdfTempfile.new("wicked_pdf.html", options[:temp_path])


unixmonkey:
Thanks a lot! :+1:

PR #451, opened 2015-09-15
Use `debug` instead of `debug=1`.

joshuapinter:
Minor adjustment and I'm not sure if you want this in there but I thought I would share it and give you the choice. It's what I'm using in my environment and it feels much nicer to just add `debug` onto the URL as opposed to `debug=1` to it.

Thanks so much, again, for creating this library. So money.


joshuapinter:
:+1:

PR #452, opened 2015-09-15
Asset URL fixes

bsedat:
1. Fixed some handling for relative asset URLs getting protocols stuck onto them
2. Fixed detection of asset URLs that weren't surrounded by quotes
3. Expanded the AssetsTest and Railtie logic for more versions of Rails. If this is controversial I can split it into a new pull request.


unixmonkey:
Awesome. I love PRs with tests that show what problem is actually being solved!

PR #454, opened 2015-09-30
Fix issue rendering stylesheets with embedded assets

codenamev:
Fix issue rendering stylesheets with embedded assets.  An example being custom fonts.


unixmonkey:
Awesome!

PR #477, opened 2015-12-29
Embed JavaScript in Base64 encoded form

thomasfedb:
Embedding JavaScript directly in HTML using `<script>` tags can fail if the embedded JS contains certain character sequences.

This can be resolved by encoding our JS as a Base64 string and then embedding it as a data URI.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -39,7 +39,7 @@ module WickedPdfHelper
     def wicked_pdf_javascript_include_tag(*sources)
       sources.collect do |source|
         source = WickedPdfHelper.add_extension(source, 'js')
-        "<script type='text/javascript'>#{read_asset(source)}</script>"
+        "<script type='text/javascript' src='#{wicked_pdf_asset_base64(source)}'></script>"
       end.join("\n").html_safe
     end
 


unixmonkey:
I don't think `wicked_pdf_asset_base64` is quite ready for this, as it doesn't cover all the edge cases that `read_asset` does, and this would break quite a few apps out there.

You can use `wicked_pdf_asset_base64` in your apps today like this:
```html
<%= stylesheet_link_tag(wicked_pdf_asset_base64('pdf.css')) %>
<%= javascript_include_tag(wicked_pdf_asset_base64('somefile.js')) %>
<%= image_tag(wicked_pdf_asset_base64('foo.png')) %>
```

As long as your asset is actually part of the asset pipeline, and can be found with `Rails.application.assets.find_asset('somefile.js')`.
The goal right now is to make `wicked_pdf_asset_base64` work with files in /public, as well as assets hosted on a CDN.

This might be as simple as swapping in `read_asset` inside `wicked_pdf_asset_base64`, but I could use help testing that out in various scenarios to make sure it doesn't break anything for those that upgrade this gem. Any help there would be very much appreciated.

thomasfedb:
Hey @unixmonkey, I think you're right that updating `wicked_pdf_asset_base64` to use `read_asset` should prevent this change from causing regressions in applications that depend on the behaviour of `read_asset`.

In fact, my first version of this patch simply wrapped `read_asset` with `Base64.encode64`, but I noticed that `wicked_pdf_asset_base64` was already present and that it would handle setting the mime type in the data URI.

I will update the PR to also change `wicked_pdf_asset_base64` to use `read_asset`.

unixmonkey:
Hey, I like where this is going, and really appreciate that you took the time out of your day to work on it, but wicked_pdf just isn't ready for this yet :/
I'm going to close this PR for now, and hopefully we can take another look at this sometime.

thomasfedb:
Not sure how wicked_pdf isn't ready. Can you elaborate?

unixmonkey:
Did you read the thread above?

thomasfedb:
I did, and replied to each of your comments at the time. Would you be happy to merge if I change `wicked_pdf_asset_base64` to use `read_asset`?

What sort of various scenarios would you want to test for?

unixmonkey:
Yes, I think that would be a much safer (and more useful) change.

Tricky scenarios that I worry about (that may not have anything to do with your proposed change):
* Very large files (IE7 doesn't support them, and IE8 limits data URIs to 32k)
* Files that read in other files (like CSS `@import` and `url('some/path/to/file.png')`)
* ERB, coffee, SASS, etc preprocessing
* Files in /public, as well as in /app/assets/ and `config.assets.paths`
* Files on CDNs
* Differing configurations of Rails 2/3/4/5 asset pipeline config options.

`wicked_pdf_asset_base64` isn't even mentioned in the README yet, because I know it doesn't handle much of this. If you can make it handle this in a more backwards-compatible way by leveraging `read_asset`, I think that'd be great.

PR #491, opened 2016-02-12
No values arguments processing

gf59ru:
Set `no_stop_slow_scripts` as `:boolean` for prevent its value passing (--no-stop-slow-scripts argument has no value)
If any `:boolean` option has `false` value, option not passes as command line argument (`make_option` method)

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -324,8 +324,8 @@ class WickedPdf
                                   :print_media_type,
                                   :disable_smart_shrinking,
                                   :use_xserver,
-                                  :no_background], '', :boolean)
-      r += make_options(options, [:no_stop_slow_scripts], '', :boolean)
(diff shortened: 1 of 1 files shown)


lib/wicked_pdf.rb
```diff
@@ -321,7 +325,7 @@ def parse_others(options)
                                   :disable_smart_shrinking,
                                   :use_xserver,
                                   :no_background], '', :boolean)
-      r += make_options(options, [:no_stop_slow_scripts], '', nil)
+      r += make_options(options, [:no_stop_slow_scripts], '', :boolean)
```
unixmonkey: Could you please put `:no_stop_slow_scripts` in the boolean array in the line above this one, then delete this line?

It should have the same result.

Thank you for your contribution!

unixmonkey:
Thanks!

PR #499, opened 2016-03-10
Removes circular require from lib/wicked_pdf/pdf_helper.rb

dummied:
In Ruby 2.3.0:

```
lib/wicked_pdf/pdf_helper.rb:2: warning: loading in progress, circular require considered harmful
<spew>
```


unixmonkey:
Yay! Thanks homie.
| [ •◡•]|

PR #503, opened 2016-03-10
Regex ambiquity

dummied:
On 2.3.0 (fairly stock):

```
warning: ambiguous first argument; put parentheses or a space even after `/' operator
```

Switched to `%r` for the offending regular expressions.


test/functional/pdf_helper_test.rb
```diff
@@ -21,7 +21,7 @@ def teardown
     test 'should prerender header and footer :template options' do
       options = @ac.send(:prerender_header_and_footer,
                          :header => { :html => { :template => 'hf.html.erb' } })
-      assert_match /^file:\/\/\/.*wicked_header_pdf.*\.html/, options[:header][:html][:url]
+      assert_match %r(^file:\/\/\/.*wicked_header_pdf.*\.html), options[:header][:html][:url]
```
janxious: Escaping `/` is not necessary with the `%r` syntax.

PR #511, opened 2016-03-19
Test refactor

dummied:
We had a new WickedPdf instance being created in `test/unit/wicked_pdf_test.rb`.

Figured we should use it throughout rather than instantiating a fresh one in many of the tests.


unixmonkey:
Sweet. Thanks!

PR #515, opened 2016-03-30
Embed JavaScript in Base64 encoded form

thomasfedb:
Follow up to #477 

I've updated `wicked_pdf_asset_base64` so that it uses `read_asset`, and moved a `thow` statement into `read_asset`.

@unixmonkey let me know if you think anything else should be changed.


thomasfedb:
Hey @unixmonkey, this PR was based on your feedback to #477. Is it likely that this can be merged at some point?

thomasfedb:
Will investigate why this is failing also!

PR #518, opened 2016-04-06
Add window_status flag

jhollinger:
`--window-status="ready"` is an alternative to `--javascript-delay=5000` useful when the delay might be highly variable. wkhtmltopdf waits until some JS sets `window.status === "ready"` before rendering the HTML. A very useful option for this gem to allow.


unixmonkey:
Very nice. Could you add that (with a commented note about "ready") to the mega-list of options in the README?

jhollinger:
Done

unixmonkey:
Thanks!

Issue #510, opened 2016-03-15
Rails 5 depreciation warnings


gregblass:
```
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from block in included at /Users/gregblass/.rvm/gems/ruby-2.2.2@project/bundler/gems/wicked_pdf-faa3c0d5f21b/lib/wicked_pdf/pdf_helper.rb:11)

DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from block in included at /Users/gregblass/.rvm/gems/ruby-2.2.2@project/bundler/gems/wicked_pdf-faa3c0d5f21b/lib/wicked_pdf/pdf_helper.rb:12)

DEPRECATION WARNING: after_filter is deprecated and will be removed in Rails 5.1. Use after_action instead. (called from block in included at /Users/gregblass/.rvm/gems/ruby-2.2.2@project/bundler/gems/wicked_pdf-faa3c0d5f21b/lib/wicked_pdf/pdf_helper.rb:13)
```

gregblass:
The after_filter -> after_action is trivial, but I'm trying to wrap my head around how you'd refactor the alias_method_chain. This article seems to be explaining it pretty thoroughly, but it's currently hurting my brain: http://www.justinweiss.com/articles/rails-5-module-number-prepend-and-the-end-of-alias-method-chain/

If someone can make sense of that, a fork/pull request would be nice to silence those every time tests are run. They need to be refactored to modules?

unixmonkey:
Here is my current plan on how to deal with this:

The only reason we hook into `render` with `alias_method_chain` in the first place is to provide a consistent-ish API with Rails's existing `render` API.

We use the existence of a `:pdf` hash key to determine wether this render call should be processed through `wicked_pdf` or not.

I think that it would be an improvement if we made wicked_pdf `render` calls more explicit, so that instead of this:
```ruby
format.pdf do
  render pdf: 'mypdf', debug: params[:debug]
end
```

You'd have a more explicit renderer method like this:
```ruby
format.pdf do
  wicked_pdf debug: params[:debug]
end
```

[...]

PR #519, opened 2016-04-13
Fix Rails5 deprecation warning

takkanm:
ref: #510 

I tried fix deprecation warning.

### DIFF
--- a/lib/wicked_pdf/pdf_helper.rb
+++ b/lib/wicked_pdf/pdf_helper.rb
@@ -19,7 +19,7 @@ module PdfHelper
     return if base != ActionController::Base
 
     base.class_eval do
-      after_filter :clean_temp_files
+      after_action :clean_temp_files
 
       alias_method :render_without_wicked_pdf, :render
       alias_method :render_to_string_without_wicked_pdf, :render_to_string
--- a/lib/wicked_pdf/pdf_helper.rb
+++ b/lib/wicked_pdf/pdf_helper.rb
@@ -13,6 +13,27 @@ module PdfHelper
     end
   end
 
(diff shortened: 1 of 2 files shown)


codeodor:
Are you not getting a bunch of these deprecations too?

```
DEPRECATION WARNING: `render :text` is deprecated because it does not actually render a `text/plain` response. Switch to `render plain: 'plain text'` to render as `text/plain`, `render html: '<strong>HTML</strong>'` to render as `text/html`, or `render body: 'raw'` to match the deprecated behavior and render with the default Content-Type, which is `text/plain`. (called from render_with_wicked_pdf at /Users/sam/.rvm/gems/ruby-2.3.0@cte/bundler/gems/wicked_pdf-8f970f29c4de/lib/wicked_pdf/pdf_helper.rb:22)
```

codeodor:
One error I get with this when I add `gem 'wicked_pdf', github: 'takkanm/wicked_pdf', branch: 'fix_warning_in_rails5'` to my `Gemfile` and run my tests is: 

```
ActionView::Template::Error: No such file or directory @ rb_sysopen - ~/myapp/public/stylesheets/application.css
    app/views/layouts/print.html.erb:5:in `_app_views_layouts_print_html_erb__3725552830765960760_70238720917040'
    ...
```

If I switch it to `gem 'wicked_pdf', github: 'mileszs/wicked_pdf'` there is no problem though.

unixmonkey:
Thanks a lot!

Carlows:
hey guys, the gem is still throwing me these deprecation warnings, and I'm using version `1.0.6`, they only happen in my test enviroment though :confused:

PR #525, opened 2016-05-09
Fix Rubocop errors and enforce in CI

toddsiegel:
Fixes the existing Rubocop errors and runs Rubucop in CI.

This should prevent future PRs like this. 😉 

Also, the `rubocop` `rake` task had a guard clause that was `return`ing, which is not allowed. This caused it to bomb on Ruby 1.9.2 in CI. I changed it to `next` which seems to be the proper way to do it.

If you want any of the Rubocop rules disabled instead, let me know. 

rubocop

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -87,7 +87,7 @@ class WickedPdf
     generated_pdf_file.rewind
     generated_pdf_file.binmode
     pdf = generated_pdf_file.read
-    fail "PDF could not be generated!\n Command Error: #{err}" if pdf && pdf.rstrip.empty?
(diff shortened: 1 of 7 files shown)


lib/wicked_pdf.rb
```diff
@@ -87,7 +87,7 @@ def pdf_from_url(url, options = {})
     generated_pdf_file.rewind
     generated_pdf_file.binmode
     pdf = generated_pdf_file.read
-    fail "PDF could not be generated!\n Command Error: #{err}" if pdf && pdf.rstrip.empty?
+    raise "PDF could not be generated!\n Command Error: #{err}" if pdf && pdf.rstrip.empty?
     pdf
```
toddsiegel: This was supposed to have been fixed in 2cfa57e331 but did not make it in for some reason.

PR #561, opened 2016-07-26
Find assets via build_environment in Sprockets 3.

dstrus:
With `config.assets.compile` set to `false` (which it is by default in production in Rails 5), `Rails.application.assets` returns nil. In that case, we can call `Sprockets::Railtie.build_environment(Rails.application).find_asset` in place of `Rails.application.assets.find_asset`.


unixmonkey:
Great! Do you think you could extract this out to a private method or something that lines 74 and 106 could also use?

dstrus:
Done!

unixmonkey:
Thanks a ton!

PR #565, opened 2016-08-19
Fix how to specify layout in README

boatrite:
Used to say just "pdf", but this is incorrect. Must specify "pdf.html"

This page has the correct documentation: https://github.com/mileszs/wicked_pdf/wiki/Configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ class ThingsController < ApplicationController
                disposition:                    'attachment',                 # default 'inline'
                template:                       'things/show',
                file:                           "#{Rails.root}/files/foo.erb"
-               layout:                         'pdf',                        # for a pdf.html.erb file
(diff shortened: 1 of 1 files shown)


unixmonkey:
Including the file extension in the template or layout name should only be necessary if using a different format in the file extension than the request came in as.

```ruby
def show
  respond_to do |format|
    format.pdf do # this tells Rails to look for a view with an extension of `pdf` in the filename
      render pdf: 'pdf', layout: 'mylayout' # renders mylayout.pdf.erb OR mylayout.html.pdf.erb for a layout that works for both PDF and HTML.
      render pdf: 'pdf', layout: 'mylayout.html' # renders mylayout.html.erb
    end
  end
end
```

Let me know if there's something better I could do to document this, but this is standard Rails behavior.

PR #568, opened 2016-08-25
Fix wkhtmltopdf path detection when HOME environment variable is unset

GUI:
If running in an environment where the HOME environment variable is not set, then the File.expand_path possibly called later on would throw a `couldn't find HOME environment -- expanding ~` error. This fixes it by only appending `~/bin` to the search path if the HOME environment variable is actually set.


unixmonkey:
Thanks!

Issue #111, opened 2012-05-03
wicked_pdf and remotipart gems can't coexist


woodwardjd:
I don't have the depth of expertise to identify if one or both projects are doing something inadvisable, so I'm posting this on both projects in the hope that someone more knowledgeable will be able to point out a fix to one or both parties.  At the very least, these gems don't seem to play well together.  

Both gems mess with the render method in ActionController::Base.  wicked_pdf does method chaining and remotipart does it by calling super.

When wicked_pdf and remotipart are both loaded in the Gemfile part of remotipart ceases to function.

See https://github.com/JangoSteve/remotipart/issues/27 for more details.

ilucin:
I have the same issue with Turbolinks. Turbolinks is also overriding render and when I add wicked_pdf gem that override stops working. Commenting out self.included (as mentioned above) works but I would like to avoid using forked library just because of this issue.

Since Turbolinks is kinda official rails lib (they are really pushing it) I would try to come up with a solution for this. I'm not a ruby developer so I can't really help much :/

Grammarella:
+1

We use the forked branch and have to avoid updating the gem due to this issue.

unixmonkey:
This should be solved with the merge of #574 

Please reopen if there are any further issues after upgrading.

pierre-alain-b:
Hi there,

I see this problem popping up again with wicked_pdf 2.0.1 and remotipart 1.4.x. If I downgrade to wicked_pdf 1.4.0, the problem disappears.

Best
Pierre

unixmonkey:
The only thing that I can see that could be causing this for you is this change: https://github.com/mileszs/wicked_pdf/compare/1.4.0..2.0.1#diff-ea9340aeb7bd0bcc5782373373980c56R9

introduced in #869 

I think maybe the deferring of loading makes it so that remotipart again get loaded first, causing render issues.

My plan for wicked_pdf version 3 is to stop mucking with render entirely, but it's a big change to the API surface.

Until then, you should be able to do something like this: https://github.com/mileszs/wicked_pdf/issues/111#issuecomment-5562812

or roll back to the working version. I do plan on working to fix this soonish, but it'll still likely be awhile before it's released unless you, or someone else, can help.

pierre-alain-b:
First thank you for the prompt reply. I agree the only change that could explain the new behavior would be there:
https://github.com/mileszs/wicked_pdf/pull/869/files#r386040878

[...]

PR #574, opened 2016-09-19
Use inheritance when prepending - Fixes conflict with other render gems

vlymar:
After upgrading to rails 5 we found that the way wicked_pdf prepends `ActionController::Base#render` causes an infinite loop when combined with another gem prepending render.

I believe our issue is generally related to https://github.com/mileszs/wicked_pdf/issues/111, though this PR focuses on prepend behavior.

The loop happens because wicked_pdf is not using the traditional prepend inheritance pattern, which looks like this

``` ruby
class Base
  def render
  end
end

class Patch
  def render
    patched_behavior
    super
  end
end

Base.prepend(Patch)
```

Instead, PdfHelper aliases the current render method away, then defines its own render method _on the base class_, which then calls the new render_with_wicked_pdf method.

If any other gem prepends its own render method using the inheritance pattern, before wicked_pdf prepends PdfHelper, this happens:
1. ActionController::Base is loaded, and its render method is defined.
2. OtherGem has its patch prepended, putting its render method in front of ActionController::Bases render method in the lookup chain. (OtherGem calls super to get to the original render method)
3. wicked_pdf has PdfHelper prepended
4. PdfHelper.prepended aliases OtherGem#render to render_without_wicked.
5. PdfHelper.prepended re-defines render on ActionController::Base using base_eval.
6. render is called from a controller action:
7. OtherGem#render is called, and it calls super
8. PdfHelper's version of #render is called, and it calls render_with_wicked
9. render_with_wicked calls render_without_wicked
10. render_without_wicked is an alias for OtherGem#render
11. infinite loop.

I've included a test that demonstrates this behavior. Its quite ugly as it has to replicate what railtie.rb does after another gem is prepended, which i couldn't figure out how to do since all of wicked is loaded before the tests run, so my solution was to just demonstrate it with a mock ActionController::Base class.

I'm new to test-unit so please let me know if these tests can be cleaned up.

### DIFF
--- a/lib/wicked_pdf/pdf_helper.rb
+++ b/lib/wicked_pdf/pdf_helper.rb
@@ -19,18 +19,15 @@ class WickedPdf
 
       base.class_eval do
         after_action :clean_temp_files
-
-        alias_method :render_without_wicked_pdf, :render
-        alias_method :render_to_string_without_wicked_pdf, :render_to_string
-
-        def render(options = nil, *args, &block)
-          render_with_wicked_pdf(options, *args, &block)
-        end
-
-        def render_to_string(options = nil, *args, &block)
-          render_to_string_with_wicked_pdf(options, *args, &block)
-        end
       end
+    end
+
+    def render(options = nil, *args, &block)
+      render_with_wicked_pdf(options, *args, &block)
+    end
+
+    def render_to_string(options = nil, *args, &block)
+      render_to_string_with_wicked_pdf(options, *args, &block)
     end
 
     def render_with_wicked_pdf(options = nil, *args, &block)
@@ -38,8 +35,12 @@ class WickedPdf
         log_pdf_creation
         options[:basic_auth] = set_basic_auth(options)
         make_and_send_pdf(options.delete(:pdf), (WickedPdf.config || {}).merge(options))
-      else
+      elsif respond_to?(:render_without_wicked_pdf)
+        # support alias_method_chain (module included)
         render_without_wicked_pdf(options, *args, &block)
+      else
+        # support inheritance (module prepended)
+        method(:render).super_method.call(options, *args, &block)
       end
     end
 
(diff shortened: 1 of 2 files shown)


unixmonkey:
Very cool. This has been an open issue for awhile with [remotipart, turbolinks, and possibly others](https://github.com/mileszs/wicked_pdf/issues/111)

My desired approach in the future would be to eliminate hooking into `render` at all (and instead calling `wicked_pdf(options)` in your controller, but this could be great for those that depend on this behavior or are upgrading.

I'll take a closer look soon. Do you think this should increment the MAJOR or MINOR value if conforming to [Semver](http://semver.org/)?

Can you forsee be any backwards compatibility concerns with this? I do see you are checking `respond_to?(:render_without_wicked_pdf)`, which looks like it would cover it, and it's only called on platforms that use `prepend`.

Thanks a ton for giving this a proper look!

vlymar:
No problem! Thanks for the prompt response :)

I think this is a PATCH level change...as far as I can tell I haven't changed any functionality.

The only thing I can think of that could break backwards compatibility is `super_method`, which only exists on ruby >= 2.2. Like you said, currently it would only ever be called with rails 5, but if down the line someone accidentally removes the alias_method_chain for rails < 5, its possible render_with_wicked_pdf will fall through to the else and call super_method on a ruby that doesn't have it... I think thats fairly far fetched though, and perhaps can be caught with a test.

Personally I like the [Renderers.add](http://apidock.com/rails/v4.2.7/ActionController/Renderers/add/class) approach the most (that feature appears to exist as early as rails 3). It would let you call `render pdf: 'whatever'` without any monkeypatching at all.
#111 describes a patch that uses add_renderer. I guess the problem there is add_renderer might not handle render_to_string, though i'm not sure what the use case for that is. At the very least you reduce the monkeypatching by 50%

unixmonkey:
I've been sitting on this excellent work for too long. Thanks a ton for the help.

parvesh-hyperiq:
Fixes are added to master branch.

PR #575, opened 2016-09-20
Use Rails logger instead of p

guguso:


lib/wicked_pdf.rb
```diff
@@ -105,7 +105,7 @@ def on_windows?
   end
 
   def print_command(cmd)
-    p '*' * 15 + cmd + '*' * 15
+    Rails.logger.debug '*' * 15 + cmd + '*' * 15 if defined?(Rails)
```
unixmonkey: The `if` shouldn't be necessary since this is only called by `in_development_mode?`, which already does this check.

This is a good change, though!

guguso:
@unixmonkey you're right for the `if`. I removed it.

unixmonkey:
Thanks!

Issue #581, opened 2016-10-07
Javascript only loads in debug mode


lbrito1:
I'm working on a dashboard application using wicked pdf. With `show_as_html` set to `true`, my javascript files included with `wicked_pdf_javascript_include_tag` load up fine:

![screenshot from 2016-10-07 15-26-48](https://cloud.githubusercontent.com/assets/2048939/19201155/def7869a-8ca2-11e6-91f8-a46c6214fc9a.png)

But when not in debug mode, i.e. when `show_as_html` = false, no javascript files are loaded. It does not seem to be a matter of if they have enough time to finish: I'm fairly certain none of them load at all (put some `console.log`s in the beginning of the files, nothing is printed in console).

The same js-loaded chart in the picture above doesn't load (note that css-defined styles loads fine, though):

![screenshot from 2016-10-07 15-26-36](https://cloud.githubusercontent.com/assets/2048939/19201156/defa8480-8ca2-11e6-81a2-29171e54ffb7.png)

Is this somehow expected behavior? Am I missing something?

lbrito1:
Update: Javascript works as expected if I use `wkhtmltopdf` to directly render the html page, bypassing wickedpdf. So the issue is probably with wickedpdf.

unixmonkey:
What does the output (from viewing the source) around the JS tags look like with `show_as_html: true`? Are they fully qualified paths or URLs?

If they are relative links, the wkhtmltopdf headless browser won't be able to reference them when rendering. You probably need to add the reference javascript files to your precompiled assets list [as shown here on the README](https://github.com/mileszs/wicked_pdf#asset-pipeline-usage).

lbrito1:
>What does the output (from viewing the source) around the JS tags look like with show_as_html: true? Are they fully qualified paths or URLs?

The actual js code is inserted just as they would using Rails' helpers. So around `<%= wicked_pdf_javascript_include_tag "dashboards/surcharge.js" %>` I have `surcharge.js` code.

The only paths I see in the source are font-related.

I had already added the relevant js files to the precompile list (except some larger libs like jQuery, d3 etc which are declared in application.js - do I also have to somehow include them with wickedpf helpers?).

unixmonkey:
Yes, any JS you need to render your PDF version.

lbrito1:
Okay, that doesn't include sprockets-handled js, right? They seem to make their way to the final html fine.

[...]

PR #583, opened 2016-10-11
Adds BOM to intermediary html file

lbrito1:
Popular js apis use UTF-8 symbols[1]. Even though UTF-8 is the default encoding for Ruby string since v2.0[2], BOM is not added to IO streams by default.

This commit adds the UTF-8 BOM marker (`\uFEFF`) to the intermediary html generated by wickedpdf. This seems to be necessary for wkhtml to properly identify the file as UTF-8. Without the BOM marker, although the rest of the file is encoded correctly, js fails on the first non-ascii symbol (see [1]) with a SyntaxError.

Related discussion: https://github.com/mileszs/wicked_pdf/issues/581

**maintainers/reviewers**: I couldn't get the tests to run (error log [here](https://gist.github.com/lbrito1/f9037061a0e4003d81799cab92da95de)) nor am I sure if this PR warrants a test or test modification. Let me know either way.
#### Notes

[1] e.g. D3 charting api https://github.com/d3/d3.github.com/blob/5295daf46336856c5d7a6d3a4818bf23a231b31e/d3.v3.js#L1261
[2] https://ruby-doc.org/core-2.2.0/Encoding.html#class-Encoding-label-Script+encoding

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -54,6 +54,7 @@ class WickedPdf
     options.merge!(WickedPdf.config) { |_key, option, _config| option }
     string_file = WickedPdfTempfile.new('wicked_pdf.html', options[:temp_path])
     string_file.binmode
+    string_file.write("\uFEFF")
     string_file.write(string)
     string_file.close
 


unixmonkey:
Closing this branch, since the issue named can be avoided by adding `<meta charset="utf-8">` to the PDF file or layout [as described in the README here](https://github.com/mileszs/wicked_pdf#super-advanced-usage).

Although I think UTF-8 is a reasonable default, I worry that this would potentially break documents that are not UTF-8 compatible, or specified with another encoding.

Thanks for the contribution, though. I didn't realize you could do this!

PR #585, opened 2016-10-25
default_protocol in list of available config

benortiz:
`default_protocol` is an available config option that is not documented anywhere. Add it to the list of available options so that others can configure WickedPdf fully.


unixmonkey:
Thanks!

PR #590, opened 2016-10-31
WIP: Deprecate `render` and `render_to_string`.

unixmonkey:
I propose we should get rid of `alias_method_chain`, `prepend`, `super` and all that stuff and use some good-old plain and explicit render helpers.

This still needs some work around replacing `render_to_string` functionality. It works with a `join` on the end, but not a true replacement, since rendering more than one pieces isn't possible because of Rails throwing a `DoubleRenderError`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ class ThingsController < ApplicationController
     respond_to do |format|
       format.html
       format.pdf do
-        render pdf: "file_name"   # Excluding ".pdf" extension.
+        wicked_pdf
       end
     end
   end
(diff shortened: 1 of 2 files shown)


jawad-shuja:
Now, on Rails 5.1.4 we get this error:
```
/Users/jawadshuja/.rvm/gems/ruby-2.4.0/gems/wicked_pdf-1.0.6/lib/wicked_pdf/pdf_helper.rb:10:in `block in included': undefined method `alias_method_chain' for ActionController::Base:Class (NoMethodError)
Did you mean?  alias_method
	from /Users/jawadshuja/.rvm/gems/ruby-2.4.0/gems/wicked_pdf-1.0.6/lib/wicked_pdf/pdf_helper.rb:9:in `class_eval'
	from /Users/jawadshuja/.rvm/gems/ruby-2.4.0/gems/wicked_pdf-1.0.6/lib/wicked_pdf/pdf_helper.rb:9:in `included'
```

unixmonkey:
Closing for now.

PR #592, opened 2016-11-02
better asset pipeline explanation

pastullo:
I believe the Asset Pipeline explanation is key to have wicked_pdf working. The readme did a poor job at explaining that you need to include the needed assets in the precompile list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,13 @@ Using wicked_pdf_helpers with asset pipeline raises `Asset names passed to helpe
 </html>
 ```
 
+### Asset Pipeline usage
+
+It's best to serve asset files from the Asset Pipeline. The correct way to handle this is to include these assets in your ```config/initializers/assets.rb```.
+```ruby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,6 +106,13 @@ Using wicked_pdf_helpers with asset pipeline raises `Asset names passed to helpe
 </html>
 ```
 
+### Asset Pipeline usage
+
+It's best to serve asset files from the Asset Pipeline. The correct way to handle this is to include these assets in your ```config/initializers/assets.rb```.
```
unixmonkey: I do think this is an improvement, but this phrase is ambiguous. How about something more like:

"It's best to precompile assets used in PDF views. This will help avoid issues when it comes to deploying, as Rails serves asset files differently between development and production (`config.assets.comple = false`), which can make it look like your PDFs work in development, but fail to load assets in production."

I'm not sure I love that either. Seems a little long and overly technical. Any suggestions would be appreciated.

unixmonkey:
Made the change with my revised language. Thanks for inspiring it, and letting us know something isn't clear in the documentation!

PR #603, opened 2016-12-21
Update README.md with examples for Super Advanced Usage

shiroginne:
Hello,

It might be obvious how to use 
```
render_to_string('templates/pdf', layout: 'pdfs/layout_pdf.html')
```
method, but I faced the same issue several times: if footer / header is used with **content** option, PDF is generated with multiple footers for every string of a general template, if there is no **layout** option. No errors, no warnings 🤔 And templates look fine and everything is fine ("I just fix some css" 😄) but PDF contains 64 pages with footers. A couple of hours later I got it: the document was an incorrect HTML document.
I've added examples and a note, I think it can be helpful


unixmonkey:
Thanks!

PR #613, opened 2017-01-26
Handle Rails module defined without Rails env

WojtekKruszewski:
Gems like rails-html-sanitizer open `Rails` module without
actually requiring Rails. Instead check if Rails.env is defined.

Similar to: https://github.com/rails/activeresource/pull/216

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -195,24 +195,26 @@ class WickedPdf
 
   def parse_header_footer(options)
     r = []
-    [:header, :footer].collect do |hf|
-      next if options[hf].blank?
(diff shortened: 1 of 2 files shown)


WojtekKruszewski:
Two failures (with rails_edge.gemfile) are unrelated to my changes.

```
Bundler could not find compatible versions for gem "arel":
  In Gemfile:
    rails was resolved to 5.1.0.alpha, which depends on
      activerecord (= 5.1.0.alpha) was resolved to 5.1.0.alpha, which depends on
        arel (~> 8.0)

```

unixmonkey:
Cool. Thank you.

PR #615, opened 2017-02-14
Assets helper Sprockets 4 compatibility fix.

martinsabo:
Pathname attr reader was removed/deprecated. Only filename is available in Sprockets 4.x.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -68,7 +68,12 @@ class WickedPdf
           end
         else
           asset = find_asset(source)
-          asset ? asset.filename : File.join(Rails.public_path, source)
+          if asset
+            # older versions need pathname, Sprockets 4 supports only filename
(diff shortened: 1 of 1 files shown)


skinnyfit:
I need this patch, but the Travis build appears to have failed because it couldn't find ARel 8.0. I've just checked & that gem is definitely live on Rubygems now, so is anyone able to retry this build?

unixmonkey:
I just reran the Travis build and it failed exactly the same way.

```
Bundler could not find compatible versions for gem "arel":
  In Gemfile:
    rails was resolved to 5.1.0.alpha, which depends on
      activerecord (= 5.1.0.alpha) was resolved to 5.1.0.alpha, which depends on
        arel (~> 8.0)
Could not find gem 'arel (~> 8.0)', which is required by gem 'activerecord (=
5.1.0.alpha)', in any of the sources.
The command "eval bundle install --verbose " failed.
```

unixmonkey:
Anyway, I like this change. It's going in. Thanks!

PR #617, opened 2017-02-21
Update README.md

hbin:
Use suffix `html.erb` for layout didn't works for me, but `pdf.erb` did.


unixmonkey:
I'm good with this change, but would really prefer you kept `pdf` instead of `application`. Too many people run into trouble trying to load their entire application layout into a PDF.

hbin:
close this due to i delete the forked repo, opened another PR #620

PR #627, opened 2017-03-11
Fix NameError: uninitialized constant Mime::JS in Rails 5.1.0 beta

ryu39:
Thank you for your nice gem.

I found Travis CI tests with Rails edge(5.1.0 beta) gemfile are failed by `NameError: uninitialized constant Mime::JS`.

In Rails 5.1.0 beta, `::Mime::JS` and other constants are removed by this [PR](https://github.com/rails/rails/pull/26746/files#diff-c5146df11f35304765e9ceebed108f57).

I fixed this problem using `::Mime[:js]`.

But in Rails 2.3, `::Mime[:js]` cannot be used because `::Mime.[]` is not defined.
So I checked `::Mime.[]` is defined or not using `respond_to?`, and if `::Mime.[]` is not defined, then I use `::Mime::JS`.

Please check my PR.
Thank you.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper.rb
@@ -23,8 +23,9 @@ class WickedPdf
 
     def wicked_pdf_javascript_src_tag(jsfile, options = {})
       jsfile = WickedPdfHelper.add_extension(jsfile, 'js')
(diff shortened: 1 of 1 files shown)


unixmonkey:
Sweet. Thanks for this!

PR #630, opened 2017-03-19
Fix Rails 5 deprecation notice with conditional calling of appropriate hook

cgeorgii:
New pull request without failing code for Rails versions > 3 (didn't test for earlier). As suggested, the correct hook is being called if defined, otherwise failing gracefully.

Thanks for the tips!


unixmonkey:
Thanks!

PR #633, opened 2017-03-21
1.1.0.leave args alone

ladelfa:
WickedPdf::PdfHelper#render_with_wicked_pdf (and render_to_string_with_wicked_pdf) erroneously assume that there will always be arguments passed to render, and they break out the first argument, assign a default value of nil if it's missing, and then pass that to render_without_wicked_pdf, on down the chain. This breaks functionality of other rendering gems, e.g. ActiveScaffold, that may rely on getting render passed with no args (i.e. args = []) instead of a single nil one (args = [nil]).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.1.0] - 2016-09-30
+### Added
+- Support Rails 5.x and Sprockets 3.x
+- Support `window_status: 'somestring'` option, to instruct wkhtmltopdf to wait until the browser `window.status` is equal to the supplied string. This can be useful to force rendering to wait [as explained quite well here](https://spin.atomicobject.com/2015/08/29/ember-app-done-loading/)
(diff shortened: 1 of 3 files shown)


unixmonkey:
Could you please check that your issue isn't already fixed on `master`, since I have merged #574, but not yet released a new gem version?

Otherwise, could you rebase and provide a little more detail for me to reproduce the issue?

ladelfa:
What's on master currently breaks our app and I'm not sure why, so I branched from 1.1.0, made this change and now things work. Unfortunately I'm really crunched for bandwidth right now, and can't spend any additional time on it, but I wanted to put this out there in case anyone else is looking for a fix where WickedPdf corrupting the args array causes "undefined method `merge' for nil:NilClass" inside 
active_scaffold (3.4.39), lib/active_scaffold/extensions/action_controller_rendering.rb:10: in `render_with_active_scaffold'

(It may just be the way we're using ActiveScaffold link_actions, and so a somewhat rare corner condition, but it seems to me to be a reasonable expectation that render_with_wicked_pdf should pass out to render_without_wicked_pdf exactly the *args it takes in.)

unixmonkey:
This is no longer needed, especially since #925 
Thank you for raising this issue.

PR #638, opened 2017-03-29
Improve documentation on `exe_path`

jankeesvw:
My server was raising this error after a gem update:

```
Bad wkhtmltopdf's path: '/home/nginx' is not a directory.
Bundler will use '/tmp/bundler/home/root' as your home directory temporarily.
/var/deploy/timiapp.com/web_head/shared/bundle/ruby/2.3.0/bin/wkhtmltopdf
```

The fix is quite easy and can be found in this StackOverflow post: http://stackoverflow.com/a/41754326/616863

```ruby
 WickedPdf.config = {
  exe_path: '#{ENV['GEM_HOME']}/bin/wkhtmltopdf'
 }
```

With this PR I propose that this is the default in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ If your wkhtmltopdf executable is not on your webserver's path, you can configur
 
 ```ruby
 WickedPdf.config = {
-  exe_path: '/usr/local/bin/wkhtmltopdf'
+  exe_path: '#{ENV['GEM_HOME']}/bin/wkhtmltopdf'
 }
 ```
 


unixmonkey:
This is because you are using a `wkhtmltopdf` binary supplied by a gem, and that gem puts a binary (or a binstub) in that directory. This isn't always the case, and I'm unsure if `ENV['GEM_HOME']` will always be set to the right Ruby in production environments.

[This method](https://github.com/mileszs/wicked_pdf/blob/master/lib/wicked_pdf.rb#L332) shows how `wicked_pdf` will try to find your `exe_path` for you automatically if you don't create the initializer, or don't supply the `exe_path` in it.

I'm all for improving the documentation, but I think this will cause more people to have issues than it solves. I think it could be listed as a commented-out alternative to a full path, or an option where it is omitted entirely. The default one generated has that line, but commented out [as seen here](https://github.com/mileszs/wicked_pdf/blob/master/generators/wicked_pdf/templates/wicked_pdf.rb#L11-L21)

Let me know if you have any other questions or suggestions!

jankeesvw:
Thanks for the great explanation @unixmonkey, hopefully this pr with your comment will help someone in the future ✨ . I'll close it for now.

PR #646, opened 2017-04-14
Add CodeClimate badge to Readme

januszm:
Looks like there's room for improvement ;) Let's remember about it.


unixmonkey:
Good point. I've got a refactor branch that brings the GPA to 3.93 at https://codeclimate.com/github/unixmonkey/wicked_pdf/ that is mostly breaking up large methods, maybe this will motivate me to get it finished.

PR #653, opened 2017-05-24
wkhtmltopdf is not necessarily a file

ylecuyer:
WKHTMLTOPDF is not necessarrily a file:

```
WickedPdf.config = {
   :wkhtmltopdf => 'xvfb-run wkhtmltopdf
}
```


lib/wicked_pdf.rb
```diff
@@ -39,7 +39,6 @@ class WickedPdf
   def initialize(wkhtmltopdf_binary_path = nil)
     @exe_path = wkhtmltopdf_binary_path || find_wkhtmltopdf_binary_path
     raise "Location of #{EXE_NAME} unknown" if @exe_path.empty?
-    raise "Bad #{EXE_NAME}'s path: #{@exe_path}" unless File.exist?(@exe_path)
     raise "#{EXE_NAME} is not executable" unless File.executable?(@exe_path)
```
unixmonkey: Won't this executable check also raise in your case?

```ruby
File.executable?('/bin/ls') # => true 
File.executable?('/bin/ls -a') => false
```

ylecuyer: I tested and you're right, I will change the PR

Zorbash:
I think that the `:wkhtmltopdf` config option is a bit unintuitive as to what it should be (path or command).

What about introducing a `:wkhtmltopdf_command` like:

```ruby
WickedPdf.config = {
   :wkhtmltopdf_command => 'xvfb-run wkhtmltopdf'
}
```

Which would have no validations? 

I also think that the current `:wkhtmltopdf` option should be renamed to `:wkhtmltopdf_path`, so that it's clear that the validations are performed against an executable file.

unixmonkey:
@Zorbash That sounds like a good compromise, though I am hesitant to rename the existing config option until it gets properly deprecated.

There's also some confusion around the `wkhtmltopdf` and `exe_path` config options, since they essentially do the same thing, maybe some cleanup here would be good.

PR #663, opened 2017-06-21
Fix method name when checking if controllers respond to `after_action`

matthewpatterson:
https://github.com/mileszs/wicked_pdf/pull/630 used `after_action` in place of `after_filter` when it's available, since `after_filter` is deprecated in Rails 5.0. However, it was checking if the base class responded to `after_action?` (with a question mark appended) by accident.


unixmonkey:
Thanks!

PR #688, opened 2017-10-02
Provide the base_path for asset lookup

PragTob:
First, thanks for wicked_pdf! :green_heart: 

Ran into this problem while upgrading sprockets on our application
a rather lengthy debugging session through sprockets and friends
ended up here showing that the asset could be found if the
base_path was supplied. Iirc it's optional and should only
help - tbh it's a bit ago so don't have all the code in my
head anymore :)

edit: also we are running this in production so at least for us it works (tm) with the newest sprockets


unixmonkey:
Thanks!

Issue #664, opened 2017-06-29
Is the wkhtmltopdf --no-images option not supported?


bitops:
Hello! We are trying to invoke WickedPdf with the `--no-images` command line flag that `wkhtmltopdf` provides. 

We tried this:

```ruby
WickedPdf.config[:no_images] = true
```

But it didn't seem to take. When we look at what flags are being passed down to `wkhtmltopdf` we don't see `--no-images`. Are we doing something wrong? Thank you!

unixmonkey:
That must have been introduced in a relatively new version of `wkhtmltopdf`.

I think adding `:images` and `:no_images` to [this block](https://github.com/mileszs/wicked_pdf/blob/master/lib/wicked_pdf.rb#L317-L327) would fix this.

PR's are welcomed, otherwise, you can (probably) work around it for now by setting it like one of these:

```ruby
WickedPdf.config[:extra] = '--no-images'
```

or

```ruby
render pdf: 'hello', extra: '--no-images'
```

Let me know how it goes, and if you'd be interested in helping add this feature or not!

unixmonkey:
Closing now that #689 is merged. Thanks!

PR #689, opened 2017-10-03
Added image and no_image parameters.

jjculber:
Following advice from issue #664, I added a unit test to verify it failed and added the recommended fix to make the test pass.


unixmonkey:
Awesome. Thanks!

Issue #714, opened 2018-01-17
Different scaling on Windows and Linux


soundasleep:
I was having an issue where rendering PDFs on Windows and on Linux would render with different sizes, even though the underlying version of wkhtmltopdf was the same.

Upon further investigation this issue [seems to lie with wkhtmltopdf and its interaction with Linux/Windows](https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2184): 

> on Linux it prints at 75 dpi (native for WebKit) while on Windows it uses the desktop's DPI (which is normally 96 dpi). You can use --zoom 0.78125 (75/96) to get the same effect on Linux.

Would it be worth adding this caveat to the README? Alternatively (my preference), to scale the 'zoom' factor provided to wkhtmltopdf by (1.0/0.78125) if the user is on Windows?

unixmonkey:
I think an addition to the README is more appropriate. That behavior is different not only by OS, but what resolution of monitor you have, and changing a default is bound to break for a great number of users.

soundasleep:
How's this? https://github.com/mileszs/wicked_pdf/pull/715

unixmonkey:
Merged. Thanks!

PR #715, opened 2018-01-17
Add caveat on different DPI scaling on Windows/Linux

soundasleep:
See https://github.com/mileszs/wicked_pdf/issues/714 and https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2184


unixmonkey:
Great! Thanks for your help!

PR #619, opened 2017-02-23
WIP: Support for using phantomjs in addition to wkhtmltopdf

erikaxel:
Initial proof of concept for using phantomjs instead of wkhtmltopdf.
Supports the following properties: page size (format), orientation, and zoom.
To use set the phantomjs exec path in properties:
`phantomjs_exe_path: '/usr/local/bin/phantomjs'
`
I have tried to minimize changes in existing code.

It should also be noted that rasterize.js is based on the official example from Phantomjs:
https://github.com/ariya/phantomjs/blob/master/examples/rasterize.js

PR #733, opened 2018-04-10
Support for using headless chrome instead of wkhtmltopdf.

erikaxel:
Hi,

This pull request is a WIP and a starter for discussion on how wicked_pdf can support headless chrome in addition to wkhtmltopdf.

This might be useful to people since wkhtmltopdf has stagnated and is using a very old version of webkit. 

We use this code in production, so it is battle tested, but probably needs some cleanup and documentation before it can be merged. 

I also previously did a version with phantomjs https://github.com/mileszs/wicked_pdf/pull/619 but since phantomjs is also abandoned, this pull request supersedes it. However, many of the discussions in the previous pull is probably also relevant here.

Any comments more than welcome!

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -37,10 +37,12 @@ class WickedPdf
   attr_accessor :binary_version
 
   def initialize(wkhtmltopdf_binary_path = nil)
-    @exe_path = wkhtmltopdf_binary_path || find_wkhtmltopdf_binary_path
-    raise "Location of #{EXE_NAME} unknown" if @exe_path.empty?
-    raise "Bad #{EXE_NAME}'s path: #{@exe_path}" unless File.exist?(@exe_path)
-    raise "#{EXE_NAME} is not executable" unless File.executable?(@exe_path)
+    unless WickedPdf.config[:use_puppeteer]
+      @exe_path = wkhtmltopdf_binary_path || find_wkhtmltopdf_binary_path
+      raise "Location of #{EXE_NAME} unknown" if @exe_path.empty?
+      raise "Bad #{EXE_NAME}'s path: #{@exe_path}" unless File.exist?(@exe_path)
+      raise "#{EXE_NAME} is not executable" unless File.executable?(@exe_path)
+    end
 
     retrieve_binary_version
   end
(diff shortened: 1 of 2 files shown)


morellan:
Hi @erikaxel, any chance you can rebase to the latest master and fix the CI problem?, it looks like the problem is here:

```
Error: test_should_call_after_action_instead_of_after_filter_when_able(PdfHelperTest): Mocha::NotInitializedError: Mocha methods cannot be used outside the context of a test
```

erikaxel:
Rebased, and checks green.

morellan:
Thanks @erikaxel, I used your solution with Elastic Beanstalk and added a new config to setup the puppeteer executablePath on launch options, you can see it here https://github.com/morellan/wicked_pdf/commit/576c0412509b816f97dc94e458570934fbdfcd62

It was easier for me to deploy it on AWS Elastic Beanstalk Amazon AMI using a custom executable path.

cseelus:
@unixmonkey Can this be merged into master? 🤔

mynameisnotbruce:
Is there any hope for this feature? We're considering using grover (which uses headless Chrome via puppeteer), but that comes with a boatload of Node dependencies which we would love to avoid.

erikaxel:
@unixmonkey Hi, I rebased this on the latest changes so it doesn't have merge conflicts anymore. PS: We are using this in production and have for the last couple of years, so it is quite stable.  However, the code is perhaps not the most beautiful so let me know if you would like to get it in shape to actually be merged into this project.

dixpac:
@erikaxel how satisfied are you with switch from the wkhtmltopdf to puppeter? Researching to migrate also.
I'm aware that headless will be easier to work with regarding; css and dependency management, what are your production experiences in terms of performance and scalability?

erikaxel:
We are very happy with this. wkhtmltopdf is using such an old version of the renderer and with puppeteer we get the latest version whenever we want. We have used two versions, first we used one where we included Chrome in our docker build and then ran puppeteer locally. This fix does this and we used it for along time. Downside is that your image becomes quite a bit larger because of Chrome. So we have recently pulled this out and is serving it as an internal microservice instead, sending HTML, running puppetter and returning the PDF.

imlunek:
I am not sure if this is the fully working compatible code. But its very disappointing that this feature is still not merged or doesn't have similar update for using other than wkhtmltopdf.

PR #739, opened 2018-05-20
Add support for webpacker stylesheets & Javascript

LeKristapino:
This helper method supports stylesheets for Rails from webpacker gem.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -1,4 +1,6 @@
 require 'open-uri'
+# If webpacker is used, need to check for version
+require 'webpacker/version' if defined?(Webpacker)
 
 class WickedPdf
   module WickedPdfHelper
@@ -29,16 +31,27 @@ class WickedPdf
 
       def wicked_pdf_stylesheet_pack_tag(*sources)
         return unless defined?(Webpacker)
-
-        if Webpacker.dev_server.running?
+        if running_in_development?
           stylesheet_pack_tag(*sources)
         else
           css_text = sources.collect do |source|
             source = WickedPdfHelper.add_extension(source, 'css')
-            source_path = asset_pack_path(source)
-            wicked_pdf_stylesheet_link_tag(root_url[0...-1] + source_path)
+            wicked_pdf_stylesheet_link_tag(webpacker_source_url(source))
           end.join("\n")
           css_text.respond_to?(:html_safe) ? css_text.html_safe : css_text
+        end
+      end
+
+      def wicked_pdf_javascript_pack_tag(*sources)
+        return unless defined?(Webpacker)
+
+        if running_in_development?
+          javascript_pack_tag(*sources)
+        else
+          sources.collect do |source|
+            source = WickedPdfHelper.add_extension(source, 'js')
+            "<script type='text/javascript'>#{read_asset(webpacker_source_url(source))}</script>"
+          end.join("\n").html_safe
         end
       end
 
@@ -144,6 +157,28 @@ class WickedPdf
         gzipper.read
       rescue Zlib::GzipFile::Error
         nil
+      end
+
+      def webpacker_source_url(source)
+        return unless defined?(Webpacker) && defined?(Webpacker::VERSION)
+        # In Webpacker 3.2.0 asset_pack_url is introduced
+        if Webpacker::VERSION >= '3.2.0'
+          asset_pack_url(source)
+        else
+          source_path = asset_pack_path(source)
+          # Remove last slash from root path
+          root_url[0...-1] + source_path
+        end
+      end
+
+      def running_in_development?
+        return unless defined?(Webpacker)
+        # :dev_server method was added in webpacker 3.0.0
+        if Webpacker.respond_to?(:dev_server)
+          Webpacker.dev_server.running?
+        else
+          Rails.env.development? || Rails.env.test?
+        end
       end
     end
   end
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -27,6 +27,21 @@ class WickedPdf
         end.html_safe
       end
 
+      def wicked_pdf_stylesheet_pack_tag(*sources)
+        return unless defined?(Webpacker)
+
+        if Webpacker.dev_server.running?
+          stylesheet_pack_tag(*sources)
+        else
+          css_text = sources.collect do |source|
+            source = WickedPdfHelper.add_extension(source, 'css')
+            source_path = asset_pack_path(source)
+            wicked_pdf_stylesheet_link_tag(root_url[0...-1] + source_path)
+          end.join("\n")
+          css_text.respond_to?(:html_safe) ? css_text.html_safe : css_text
+        end
+      end
+
       def wicked_pdf_image_tag(img, options = {})
         image_tag wicked_pdf_asset_path(img), options
       end
--- a/README.md
+++ b/README.md
@@ -85,7 +85,6 @@ The wkhtmltopdf binary is run outside of your Rails application; therefore, your
   </body>
 </html>
 ```
-
 Using wicked_pdf_helpers with asset pipeline raises `Asset names passed to helpers should not include the "/assets/" prefix.` error. To work around this, you can use `wicked_pdf_asset_base64` with the normal Rails helpers, but be aware that this will base64 encode your content and inline it in the page. This is very quick for small assets, but large ones can take a long time.
 
 ```html
@@ -107,6 +106,13 @@ Using wicked_pdf_helpers with asset pipeline raises `Asset names passed to helpe
   </body>
 </html>
 ```
+
+#### Webpacker usage
+
+wicked_pdf supports webpack stylesheets and javascript assets.
+
+Use `wicked_pdf_stylesheet_pack_tag` for stylesheets
+Use `wicked_pdf_javascript_pack_tag` for javascripts
 
 #### Asset pipeline usage
 


lib/wicked_pdf/wicked_pdf_helper/assets.rb
```diff
@@ -27,6 +27,21 @@ def wicked_pdf_stylesheet_link_tag(*sources)
...
+      def wicked_pdf_stylesheet_pack_tag(*sources)
+        return unless defined?(Webpacker)
+
+        if Webpacker.dev_server.running?
+          stylesheet_pack_tag(*sources)
+        else
+          css_text = sources.collect do |source|
+            source = WickedPdfHelper.add_extension(source, 'css')
+            source_path = asset_pack_path(source)
+            wicked_pdf_stylesheet_link_tag(root_url[0...-1] + source_path)
```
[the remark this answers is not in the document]
LeKristapino: yes. root_url ends with a trailing slash.

lib/wicked_pdf/wicked_pdf_helper/assets.rb
```diff
@@ -27,6 +27,21 @@ def wicked_pdf_stylesheet_link_tag(*sources)
         end.html_safe
       end
 
+      def wicked_pdf_stylesheet_pack_tag(*sources)
```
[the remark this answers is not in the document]
LeKristapino: I will try to add javascript pack tag as well, though it works a bit differently than stylersheet pack tag. I don't know about asset_tag though, but will look it up and update this PR :)

LeKristapino:
Ok, will add javascript pack support and update README :)

unixmonkey:
BTW, if you rebase off of `master` the Travis tests should pass now. I had a test dependency change that I fixed last week.

lib/wicked_pdf/wicked_pdf_helper/assets.rb
aki77: It occurs error when request spec.
I think that is better in the following code.

```ruby
wicked_pdf_stylesheet_link_tag(Rails.public_path.join(source_path))
```

LeKristapino: I believe this is a riskier solution, since it uses filesystem instead of URLs. I believe that direct links to filesystem should be avoided as much as possible

LeKristapino:
@unixmonkey Updates made, waiting for a review. Maybe @damienlethiec could check if everything works now

damienlethiec:
@LeKristapino I have just tried the update and same result unfortunately :(

LeKristapino:
@damienlethiec What was the error message?

lib/wicked_pdf/wicked_pdf_helper/assets.rb
```diff
@@ -27,6 +29,32 @@ def wicked_pdf_stylesheet_link_tag(*sources)
...
+        return unless defined?(Webpacker)
+        if running_in_development?
+          stylesheet_pack_tag(*sources)
+        else
+          css_text = sources.collect do |source|
+            source = WickedPdfHelper.add_extension(source, 'css')
+            wicked_pdf_stylesheet_link_tag(webpacker_source_url(source))
```
chadmetcalf: Just dropping in, so these questions are probably more ignorance:

1 • The JS pack tag uses `#read_asset` to resolve the url/path, why does that not apply here?
2 • In non dev-server environments, should be both the stylesheet and js tags use a path helper instead of a url helper?

chadmetcalf: RE #2: It looks like `#webpacker_source_url` does some of this work.

RudyOnRails:
Is this PR required to make wicked_pdf work with Bootstrap CSS? If so, I would be happy to try and help out!

Mar1cX:
Any progress on this issue?

LeKristapino:
Sorry for leaving this issue hanging. I tried to replicate @damienlethiec issues, but unsuccessfully. I would really appreciate more feedback on what is working with this solution and what isn't - emphasis on *isn't*.

My problem is that I have changed projects and no longer work on a project using wicked_pdf. This is why testing is a bit harder, but I can still try and finish this, if there is some feedback on what is / is not working. I suppose @RudyOnRails and @Mar1cX  could help with that?

unixmonkey:
Thanks for reporting local success, but can anyone confirm this works in a production environment?

LeKristapino:
Well, One project has it running in production environment and there doesn't seem to be any problems.  But probably someone else should confirmt this :)

cseelus:
Weird: I gave this PR another try in a different project and again it didn't work (dev, didn't even test production environment). I saw that the included CSS pack uses `media="screen"` so I changed that to `media="all` and now it works. 

So @LeKristapino, any chance you could change that?

tloudon:
Thanks for this @LeKristapino 

@unixmonkey I'm running the PR in a near production environment.  We're doing final qa w/ the client before launching (it's in RAILS_ENV=production).  Our pdfs are pretty simple, but we are using webpack for bootstrap (and some other stuff not used on the pdfs).  It looks fine to me: the stylesheets are definitely included.  I'm happy to report back in a few weeks if that would be helpful.

I rebased master, if anyone wants to clone to test; https://github.com/Loudon-Company/wicked_pdf/tree/webpack .  I didn't make any other changes.

Additionally, is there a way to debug the pdf?  I've got an html version of a page that uses one layout and the pdf version uses another; so there's not much point in comparing those.  I'm not seeing the intermediate html before wkhtmltopdf does its thing.  I might be able to give more debug info if you can help me understand the internals a little more, I don't think I'll have time to dig through things on my own.

Thanks!

unixmonkey:
@tloudon You should be able to inspect the HTML that would be passed to `wkhtmltopdf` by using the `show_as_pdf` option like this:

```
format.pdf do
  render pdf: 'mypdf', show_as_html: true
end
```

Even though you are requesting a PDF, you will be served HTML that you can view-source on.

mattrothstein:
@unixmonkey Is [this](https://travis-ci.org/mileszs/wicked_pdf/jobs/395371700) rubocop issue holding this PR up? Or does something else need to be done here?

unixmonkey:
@mattrothstein The Rubocop thing isn't an issue (I mean, I guess it is, but I can take care of that myself easily enough), it's mostly that I don't have a ton of confidence this works in a production environment yet.

@cseelus was having trouble (is this still true?), and @tloudon said he'd report back and hasn't yet.

I've also been meaning to try this out myself, but haven't yet. Do you think you could help?

tloudon:
@unixmonkey sorry, should have set a reminder.  yeah, no issues w/ it in prod for us over the past couple of months.

mattrothstein:
@tloudon Did you use the javascript pack tag a swell or just using this for the stylesheets?

tloudon:
@mattrothstein hey, here's a bit of a brain dump.  Hopefully the extra info serves as helpful context not just ramblings :)

I didn't have a ton of time to work on this particular feature (PDFs in general) and ran into a lot of snags w/ fonts; format, weight, etc (I don't recall issues w/ the pack tags).  I found it hard to debug b/c even when something would look right as html w/ `show_as_html: true`, the PDF wouldn't look right (ie, wkhtmltopdf itself issues).  TL;DR, I tried a bunch of stuff and am not sure if I followed best-practices but this works for me.  (The whole font experience kind of turned my brain to mush wrt PDFs.)

```
    %style{ type: "text/css"}
      != "@font-face { font-family: 'Helvetica-Light'; font-weight: 300; src: url('#{Rails.root.join('app', 'assets', 'fonts', 'Helvetica Light.otf')}') format('opentype') }"
    %style{ type: "text/css"}
      != "@font-face { font-family: 'Helvetica-Bold'; font-weight: 700; src: url('#{Rails.root.join('app', 'assets', 'fonts', 'Helvetica Bold.otf')}') format('opentype') }"
    = wicked_pdf_stylesheet_pack_tag('application')
    = wicked_pdf_stylesheet_link_tag('pdf')
```

Note, we ended up using both webpack and the asset pipeline (wanted to get rid of the asset pipeline but some gems used it); so bootstrap and all of our front end libs are coming in via webpack and the custom sass is in the pipeline.  I don't think you would need to do this--it was just a good boundary for our project.  We had issues w/ font rendering if woff or ttf--so otf ftw.  Client is a Communications firm that started out a graphic design firm--they are happy w/ the end results now, so yeah not super smooth but overall they're happy so we're happy.

LeKristapino:
@unixmonkey I rebased this PR, so it is up-to-date with latest code. But I still don't understand whether I need to fix or improve something, because of mixed reviews :D
Seems that this feature is somewhat requested. Maybe we could speed up the process of review for this somehow?

jwcooper:
We've been using this in production for about 6 months without issue (updating to the re-based version today). Our js is managed by webpacker, css by sprockets. We have a fairly complex pdf generation setup as well, generating in real-time as well as in sidekiq workers. No issues thus far.

martinbjeldbak:
Just chipping in our successful experience with this change in support of this PR.

We have also been using @LeKristapino's change in production for a couple of months now with no issues. In our application, all assets are managed by webpacker, including css and images. We are also generating PDFs in Sidekiq and have not experienced any difficulties. Hundreds of PDFs have been rendered this way with styles and JS coming from our webpacker installation.

unixmonkey:
Thank you @LeKristapino, and thank you everyone else for chipping in with your input and production experiences on this. That gives me a lot more confidence in this, even though I haven't used this type of Webpacker  integration myself yet. Merging. I'll make some minor changes (Changelog & RuboCop changes to get the build passing again) and cut a release today or tomorrow.

unixmonkey:
This is in the released version 1.3.0. Please let me know if you have any issues with it.

Menelle:
does anyone succeed with this with webpacker 4 and rails 6.0.0rc1 ? I got everything on html version and lose styling as soon as I switch to pdf.

BookOfGreg:
I've found that because Webpacker 4 does code splitting by default, this set of helpers no longer can find the CSS. Having Webmock in my tests highlights this:
```
MyControllerTest#test_should_create_pdf:
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET http://www.example.com/packs-test/css/pdf-4d6232f2.chunk.css
```

zealot128:
Just want to chime in, had the problem that in production the css/js was not loaded, because the server itself was proxied, so a direct http call to "mydomain.com/packs/..." did not work (could maybe work when rerouting dns to the load balancer etc). 

So I build this kind of wrapper tag, which just reads the js/css directly from hard disk without http interaction. Should be fast, too.


```ruby
# some helper module
    def production_pdf_pack_tag(asset)
      file_path = 'public' + asset_pack_path(asset)
      content = File.read(file_path)
      if asset.include?('.js')
        "<script type='text/javascript'>#{content}</script>".html_safe
      else
        "<style type='text/css'>#{content}</style>".html_safe
      end
    end
```

```slim
/ view

= production_pdf_pack_tag "recruiter.js"
= production_pdf_pack_tag 'recruiter.css'
```

jasonadkison:
> Just want to chime in, had the problem that in production the css/js was not loaded, because the server itself was proxied, so a direct http call to "mydomain.com/packs/..." did not work (could maybe work when rerouting dns to the load balancer etc).
> 
> So I build this kind of wrapper tag, which just reads the js/css directly from hard disk without http interaction. Should be fast, too.
> 
> ```ruby
> # some helper module
>     def production_pdf_pack_tag(asset)
>       file_path = 'public' + asset_pack_path(asset)
>       content = File.read(file_path)
>       if asset.include?('.js')
>         "<script type='text/javascript'>#{content}</script>".html_safe
>       else
>         "<style type='text/css'>#{content}</style>".html_safe
>       end
>     end
> ```
> 
> ```
> / view
> 
> = production_pdf_pack_tag "recruiter.js"
> = production_pdf_pack_tag 'recruiter.css'
> ```

Thanks for this! I tried everything I could think of to get wicked_pdf_stylesheet_pack_tag to work in production.

PR #746, opened 2018-06-12
Add help documentation for adding to email

kieranklaassen:
It took me a while to figure this out and the article in the wiki is outdated. I think this is a use case people will find useful to read about on the readme.


unixmonkey:
Thanks!

PR #751, opened 2018-07-09
Raise exception on all pdf generation errors

thomasbarton:
Issue: When image assets fail to download from the web (return non 200 statuses).  We want to throw an exception.

Passing `extra: '--load-media-error-handling abort'` to WickedPdf should fail the pdf generation.  But it doesnt.  This seems to be because the underlying library `wkhtmltopdf` doesnt return the correct exit code it only returns error output.

Solution: enable a flag in wicked_pdf that will raise an exception if there is any output to stderr.
```
raise_on_all_errors: true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ class ThingsController < ApplicationController
                no_background:                  true,
                viewport_size:                  'TEXT',                    # available only with use_xserver or patched QT
                extra:                          '',                        # directly inserted into the command to wkhtmltopdf
+               raise_on_all_errors:            nil,                       # raise error for any stderr output.  Such as missing media, image assets 
                outline: {   outline:           true,
                             outline_depth:     LEVEL },
                margin:  {   top:               SIZE,                     # default 10 (mm)
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -85,6 +85,7 @@ class WickedPdf
     generated_pdf_file.rewind
     generated_pdf_file.binmode
     pdf = generated_pdf_file.read
+    raise "Error generating PDF\n Command Error: #{err}" if options[:raise_on_all_errors] && !err.empty?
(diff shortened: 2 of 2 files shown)


unixmonkey:
Great idea. Thanks!

thomasbarton:
Thank you for the quick merge @unixmonkey 👍 

When does wicked_pdf plan on releasing a new version of the gem?

ximus:
am I crazy or does wkhtmltopdf always output via stderr? even when there is no error:

```
>wkhtmltopdf --page-size Letter file:////foo.html bar.pdf
Loading pages (1/6)
Counting pages (2/6)
Resolving links (4/6)
Loading headers and footers (5/6)
Printing pages (6/6)
Done
```
status code is 0 and pdf is generated correctly

```
> wkhtmltopdf --page-size Letter file:////foo.html bar.pdf > out 2>error
> cat out
> cat error
Loading pages (1/6)
Counting pages (2/6)
Resolving links (4/6)
Loading headers and footers (5/6)
Printing pages (6/6)
Done
```

joshuapinter:
> am I crazy or does wkhtmltopdf (0.12.5) always output via stderr? even when there is no error:
> 
> ```
> >wkhtmltopdf --page-size Letter file:////foo.html bar.pdf
> Loading pages (1/6)
> Counting pages (2/6)
> Resolving links (4/6)
> Loading headers and footers (5/6)
> Printing pages (6/6)
> Done
> ```
> 
> status code is 0 and pdf is generated correctly
> 
> ```
> > wkhtmltopdf --page-size Letter file:////foo.html bar.pdf > out 2>error
> > cat out
> > cat error
> Loading pages (1/6)
> Counting pages (2/6)
> Resolving links (4/6)
> Loading headers and footers (5/6)
> Printing pages (6/6)
> Done
> ```

I think you're right. I'm on wkthmltopdf 0.12.3 and stdout gets used for "Loading pages", "Counting pages", etc. That makes setting `raise_on_all_errors: true` useless.

PR #764, opened 2018-09-08
WIP: Gem binary locations

tilsammans:
Hi there! A heads up!
I've been working on a patch that automatically discovers if any of the wkhtmltopdf-binary gems are installed. Along the way I made changes to the way the executable is discovered.
I think it's an improvement. At this stage I would like to know what you think.
It might be close to done or not quite yet. I'm ready for any scenario. ;)


unixmonkey:
I love the idea of this. It's been turning out to be a bigger problem lately, and I've been instructing some people to start using a variation of this:

```ruby
WickedPdf.config = {
  exe_path: "#{ENV['GEM_HOME']}/gems/wkhtmltopdf-binary-#{Gem.loaded_specs['wkhtmltopdf-binary'].version}/bin/wkhtmltopdf_linux_amd64"
}
```

depending on which gem they are using. The binary doesn't always work when called from a binstub because the stub could be loading a different environment.

Thank you for looking into this. ❤️ Let me know if you have any questions or need help!

PR #768, opened 2018-09-19
Add support for multiple documents

bibstha:
`wkhtmltopdf` supports downloading multiple html documents and combining them into a single pdf.
WickedPdf doesn't support it right now.

This PR adds this support. It introduces two new interfaces.

* `WickedPdf#pdf_from_multiple_strings`
* `WickedPdf#pdf_from_multiple_html_files`

While keeping the existing interfaces as it is.


bibstha:
Looking at the test failures, seems like we need a patched version of QT for this. 🤔

unixmonkey:
Great! This seems very useful, especially for cover pages etc!
I think you might be able to make CI use a QT patched version by modifying [this line](https://github.com/mileszs/wicked_pdf/blob/master/.travis.yml#L22)

But it'll likely need to be a bit more involved to be something like this:
```shell
sudo apt-get install -y libqt5webkit5-dev libqt5xmlpatterns5-dev libqt5svg5-dev &&
  wget https://downloads.wkhtmltopdf.org/0.12/0.12.5/wkhtmltox_0.12.5-1.bionic_amd64.deb &&
  sudo dpkg -i wkhtmltox_0.12.5-1.bionic_amd64.deb &&
  sudo apt-get install -f
```

BTW, there's a lot of cool stuff in the `wkhtmltopdf` [Travis config](https://github.com/wkhtmltopdf/wkhtmltopdf/blob/master/.travis.yml) that might help.

bibstha:
@unixmonkey I've updated the PR by adding wkhtmltopdf with patched-qt.
All the tests passed except the with ruby version 1.8.7. This is because I added pdf-inspector as a development dependency - to make sure the pdfs generated could be tested easily.
Do we still want to support 1.8.7 (Or can we move it to the "Allowed failures" section on Travis)?
Otherwise the rest of the things on this PR looks good.

unixmonkey:
I think it's high time to drop support for 1.8.7. If someone still on that platform needs this, I think it's safe to assume they are ok pinning to an existing version or writing their own patches. Doing this officially should be a major version bump though.

Since this is a development-only dependency requirement right now, let's move 1.8.7 to allowed failures.

miguelms95:
Hi! Is this cool feature going to be merged??

KristenWilde:
Would love to see this feature added.

PR #796, opened 2019-02-05
Add instructions for usage with Rails in API mode

ameft:
I had the scratch my head a lot to find the solution in the scenario.
This SO answers helped a lot:
https://stackoverflow.com/a/41050206
https://stackoverflow.com/a/51145337
https://stackoverflow.com/a/31842487

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,7 +302,13 @@ If you need to display utf encoded characters, add this to your pdf views or lay
 ```html
 <meta charset="utf-8" />
 ```
-
+If you need to return a PDF in a controller with Rails in API mode:
+```ruby
+# partial in app/views/controller_name/_action_name.pdf.erb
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -302,7 +302,13 @@ If you need to display utf encoded characters, add this to your pdf views or lay
 ```html
 <meta charset="utf-8" />
 ```
-
+If you need to return a PDF in a controller with Rails in API mode:
+```ruby
+# teamplate in `app/views/controller_name/action_name.html.erb` and layout in `app/views/layouts/pdf.html.erb`
```
unixmonkey: Typo here: teamplate => template, and I think this comment isn't necessary.

ameft: Ok thanks

PR #800, opened 2019-02-22
Add ruby 2.5, 2.6 to test on rails edge

iggant:


unixmonkey:
@iggant The PR and commit are labeled 2.5, but the code change uses Ruby 2.6.

Did you mean for this to be Ruby 2.5 or Ruby 2.6?

iggant:
@unixmonkey fixed title 
can I use apprisal gem for generate gemsets? I probably will create another PR for this.

unixmonkey:
I wouldn't be against it, but looking through the README, I don't see the advantage. Can you help explain?

Even though you've added some new gemfiles for Travis to test, these tests aren't running, because the new Rails app generation process stops when it can't find `spring`.

Are you planning to try to get things running and (and possibly passing)?

Issue #797, opened 2019-02-18
Is it possible to render an inline template to PDF string?


yowzadave:
Rails allows you to render an "inline" template, for cases where you have a dynamic template stored in a database: 

```ruby
render :inline => "<p>Hello World</p>", :layout => "application"
```

Is it possible to use `render_to_string` in the same way, by passing the complete template string, rather than the filename of a template on disk?

unixmonkey:
Yes.

Here's an example of converting an html string to a PDF and rendering it:

```ruby
def show
  pdf = WickedPdf.new.pdf_from_string(
    '<!doctype html><html><head></head><body><h1>Hello World!</h1></body></html>'
  )
  respond_to do |format|
    format.pdf end
      send_data pdf, type: 'application/pdf', disposition: 'inline'
    end
  end
end
```

Just make sure it's a complete, fully-formed HTML document, including doctype.

Please let me know if you have any trouble with this.

yowzadave:
Thank you! So in my case, if I want to use rails templating to build my HTML document, then I'm guessing I would need to use the Rails' `render_to_string` function first (not to be confused with wicked_pdf's `render_to_string` function)?

unixmonkey:
I'm not sure I'm understanding your follow-up question correctly. Maybe an example of what you are trying to do would help?

If you were using Rails' templating, then you'd use WickedPdf's hooks into `render(pdf: 'mypdf')` or `render_to_string(pdf: 'mypdf')` with the appropriate options shown in the README, which pretty much do what I just showed you under the hood, after assembling your templates and layout into an HTML tempfile.

So if you had a `PostsController#show` action, you'd create an `app/views/posts/show.pdf.erb` file with your html you want to be converted to PDF, and an `app/views/layouts/pdf.erb` file for the layout, and it'd look something like this:
```ruby
class PostsController < ApplicationController
  def show
    @post = Post.find(params[:id])
    respond_to do |format|
      format.pdf do
        render pdf: 'mypdf'
      end
    end
  end
end
```

Rails' `render_to_string` works pretty much the same as `render`, but it outputs a string instead of a response body. If you use `render_to_string(pdf: 'mypdf')`, you'll get a stringified version of the PDF that can be used to perhaps save the final PDF to your database, or as a file attachment to an email.

Here's an example of how that might be used:
```ruby
class PostsController < ApplicationController
  def email_post
    @post = Post.find(params[:id])
[...]

PR #801, opened 2019-03-06
respect rails 5 `assigns` render option

dleavitt:
Rails 5 supports an `assigns` option to `render`/`render_to_string` that allows setting instance variables available to the template. This is particularly handy for rendering outside of controllers where these instance variables wouldn't otherwise be available.

It's a pass-through so I didn't add any new tests (but could.)


unixmonkey:
Thank you very much. There are other options that probably need passed through (like `:inline` from #797), or perhaps we could take a reverse approach and only peel off options we need to mess with (as suggested in #633), but this does the job great. Thank you!

dleavitt:
thanks!

PR #812, opened 2019-03-18
Extract and encapsulate controller code

adzap:
I'm having some issues with the latest release v1.2.0. In diagnosing I read through the gem code and saw quite a bit of method pollution in the base controller I'd like to get rid of. 

I've extract the controller code from PdfHelper into a renderer class to encapsulate it. This helps remove method pollution and gives more freedom to refactor.

This PR is first of probably a few PRs to help minimise impact on base controller render path.

### DIFF
--- a/lib/wicked_pdf/pdf_helper.rb
+++ b/lib/wicked_pdf/pdf_helper.rb
@@ -36,7 +36,8 @@ class WickedPdf
 
     def render_with_wicked_pdf(options = nil, *args, &block)
       if options.is_a?(Hash) && options.key?(:pdf)
-        WickedPdf::Renderer.new(self).render(options)
+        @_pdf_renderer = WickedPdf::Renderer.new(self)
+        @_pdf_renderer.render(options)
       elsif respond_to?(:render_without_wicked_pdf)
         # support alias_method_chain (module included)
         render_without_wicked_pdf(options, *args, &block)
@@ -48,7 +49,8 @@ class WickedPdf
 
     def render_to_string_with_wicked_pdf(options = nil, *args, &block)
       if options.is_a?(Hash) && options.key?(:pdf)
-        WickedPdf::Renderer.new(self).render_to_string(options)
+        @_pdf_renderer = WickedPdf::Renderer.new(self)
+        @_pdf_renderer.render_to_string(options)
       elsif respond_to?(:render_to_string_without_wicked_pdf)
         # support alias_method_chain (module included)
         render_to_string_without_wicked_pdf(options, *args, &block)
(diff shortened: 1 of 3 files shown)


adzap:
Split out the PdfHelper into separate modules to explicit including or prepending depending on prepend support. Using the Renderer class makes this much cleaner.

I think this should help fix #810

adzap:
The build is erroring and I don't know what to do about it. Gave it a kick. See error message https://travis-ci.org/mileszs/wicked_pdf/jobs/509342770

unixmonkey:
@adzap I was able to restart all the builds that failed due to the system updates, but 1.8.7 is still failing with:
```
uninitialized constant PdfHelper (NameError)
	from /home/travis/build/mileszs/wicked_pdf/lib/wicked_pdf/railtie.rb:36
```

Thanks for alerting me to it.

I haven't stopped by to comment yet on any of your PRs yet, but I want you to know I'm very happy to see this cleanup happen.

However, I'd really like to ship a focused fix for the remotipart issue separately if possible, then we can roll these up into the next minor or perhaps major release.

If you could help with that, I'd be really grateful, but for now I think I'm just going to try and reverse the changes from #574, which I think should make 1.2 a stable upgrade for 1.1 users.

adzap:
@unixmonkey I think that is a great plan. I'll let you get back to getting a stable release out. 

I'd be very happy to help get a major release out. It think it would help to drop some old Rails versions and old Ruby versions to help clean up the gem quite a bit.

Let me know when you want to proceed. I am happy to work on a branch to support this.

unixmonkey:
My plans for breaking changes that would warrant a major release (could be one or several), would be to:

1. Create a transitionary release where stuff that's going away throws deprecation warnings.
2. Add ability to opt-out of, or opt-in to hooking into render with `render pdf: 'mypdf'`, so you can include it yourself only where desired & call something like `wicked_pdf(options)` directly.
3. Remove support for at least 1.8 (probably 1.9 too), and embrace newer hash syntax.
4. Remove support for Rails 2 & 3
5. Option to or separate helpers for `wicked_pdf_stylesheet_link_tag` and `wicked_pdf_javascript_include_tag` to opt-in or out of dumping content to a style or script tag. This behavior optimizes rendering speed in some cases, but borks some libraries.

New stuff that I'd really like to give attention, but not related to breaking changes (that I know of yet), better support with Webpacker and ActiveStorage configurations.

adzap:
Sounds Good. 

My focus was to initially clean up the code and removing it from base Rails classes as much as possible. This would also include some work on extracting the helper code into a classes to allow it to be worked on more freely. I'm still on Rails 4.5 for my work so my webpacker experience is nil. 

As for your point 2, the renderer class and the railtie changes to add a format specific renderer block that takes care of the issues with overriding render. You could make it opt in also with a combination of config and an after_initialize block to add it late.

So far my PRs are non-breaking of the public API, or could be made so (I can clean up this PR to make it non-breaking with by using a render method override always).

So I'm not sure if you are comfortable with merging some of my PRs for a minor release.

PR #813, opened 2019-03-18
Use module#prepend when available.

unixmonkey:
In #810 It was discovered that version `1.2.0` was released with an issue that would result in a `SystemStackError (stack level too deep)`.

I believe this has to do with the usage of `alias_method_chain` applied to `ActionController::Base#render`, which may also be tampered with by something else in the `Gemfile`.

### DIFF
--- a/lib/wicked_pdf/railtie.rb
+++ b/lib/wicked_pdf/railtie.rb
@@ -5,20 +5,16 @@ require 'wicked_pdf/wicked_pdf_helper/assets'
 class WickedPdf
   if defined?(Rails.env)
 
-    if Rails::VERSION::MAJOR >= 5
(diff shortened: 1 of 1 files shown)


lib/wicked_pdf/railtie.rb
```diff
@@ -5,20 +5,16 @@
 class WickedPdf
   if defined?(Rails.env)
 
-    if Rails::VERSION::MAJOR >= 5
+    if Rails::VERSION::MAJOR >= 4
```
unixmonkey: The only difference between the Rails 4 and 5+ paths was `:include` vs `prepend`, so they were consolidated, and feature detection is now being used to determine whether to use `:prepend` or `:include`.

PR #814, opened 2019-03-19
Adjust gemspec development dependencies

adzap:
When running the tests for local dev for the first time I ran into a few issues. 

Firstly the bundler version in Ruby >= 2.5 was pegged to bundler >= 2. I don't think this necessary as I am happily continuing to use bundler v1.17.x for the moment and didn't want to be forced to upgrade. So I have relaxed the bundler version to use what is currently installed since there is no dependency on v2 in the core code.

Also added bootsnap since the dummy app requires it when booting.

### DIFF
--- a/wicked_pdf.gemspec
+++ b/wicked_pdf.gemspec
@@ -29,7 +29,7 @@ DESC
   spec.add_dependency 'activesupport'
 
   spec.add_development_dependency 'rails'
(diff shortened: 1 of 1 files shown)


unixmonkey:
I had to fix some merge conflicts, and since you authored this PR, the `bundler` version clamps have been removed entirely, but addtion of `bootsnap` for default development is a good one.

Thank you!

PR #816, opened 2019-03-20
Encapsulate binary path and version handling

adzap:
Cleans up main class and allows refactoring of path handling.

Binary path discovery can be complex it seems and this offers a path to improved handling or overriding if necessary. 

Long term goal of combining extracted option parsing and this to allow a command class to combine and execute and return result.

I think these combined PRs provide a path to a well factored and a cleaner handling gem's functions.

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -26,26 +26,24 @@ end
 require 'wicked_pdf/version'
 require 'wicked_pdf/railtie'
 require 'wicked_pdf/tempfile'
+require 'wicked_pdf/binary'
 require 'wicked_pdf/middleware'
 require 'wicked_pdf/progress'
 
 class WickedPdf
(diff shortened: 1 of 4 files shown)


adzap:
Same problem with the build again.

unixmonkey:
Rebuild passed :)

adzap:
thanks! 

The overall ambition, to be able to combine an option parser object with a binary object consumed by a command object which handles the actual cli execution. This is coordinated by a document object that is home to pdf_from_string, pdf_from_url etc. methods.

Decoupled, testable, customisable.

adzap:
This redesign would support things like #768 much more easily and cleanly.

unixmonkey:
I had to resolve some minor merge conflicts related to `xvfb`, but this is now merged.

Thanks!

PR #832, opened 2019-04-30
Add note about options based on wkhtmltopdf version.

### LINKED ISSUE #478: Unknown long argument --toc-depth
When using `toc: { depth: 3 }` as the [README documentation](https://github.com/mileszs/wicked_pdf#advanced-usage-with-all-available-options) shows, it produces the following error from `wkhtmltopdf`:

``` bash
Command Error: Unknown long argument --toc-depth
```

This is running on version 0.12.2.1 of `wkhtmltopdf`.

It looks like that option is no longer supported in `wkhtmltopdf` or it may have been renamed to `--outline-depth` as per http://wkhtmltopdf.org/usage/wkhtmltopdf.txt.

joshuapinter:
I would much prefer to label each option with the version of wkhtmltopdf that is required to use it, and possibly commands for older or newer versions, but I don't have to time to verify that and will be more difficult to maintain. Ultimately, the source of truth should be the wkhtmltopdf docs of your specific version. But this note at least warns the user about this requirement.

Fixes #478.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,9 @@ In this case, you can use that standard Rails helpers and point to the current C
 ```
 
 ### Advanced Usage with all available options
+
+_NOTE: Certain options are only supported in specific versions of wkhtmltopdf._
+
 ```ruby
 class ThingsController < ApplicationController
   def show


unixmonkey:
Thanks!

joshuapinter:
👍 Thank YOU! :)

PR #834, opened 2019-05-20
Add support for --log-level option

eizengan:
Without the ability to override it, `--log-level info` is implied when `wkhtmltopdf` executes. This - for some unholy reason - sends progress messages to stderr, which has the unfortunate side-effect of making WickedPdf panic dramatically over successfully completed tasks when run with `raise_on_all_errors: true`.

The intent of the changes herein is to ensure sensible behavior when running with `log_level: 'error', raise_on_all_errors: true`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ class ThingsController < ApplicationController
                viewport_size:                  'TEXT',                    # available only with use_xserver or patched QT
                extra:                          '',                        # directly inserted into the command to wkhtmltopdf
                raise_on_all_errors:            nil,                       # raise error for any stderr output.  Such as missing media, image assets
+               log_level:                      'none, error, warn, info',
                outline: {   outline:           true,
                             outline_depth:     LEVEL },
                margin:  {   top:               SIZE,                     # default 10 (mm)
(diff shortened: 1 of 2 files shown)


unixmonkey:
Oh, that's odd. I wonder if it's the same for older versions of `wkhtmltopdf`?

We used to pass `-q` (or `--quiet`) on non-Windows systems by default, but that got changed at some point somewhat recently. I wonder if that fixes your issue, or causes it to ignore errors?

Would you mind checking that out and maybe adding an option for `--quiet` too?

eizengan:
@unixmonkey Ah! Very likely related. From the man page:
> [-q is] maintained for backwards compatibility; Same as using --log-level none
I can add support for `--quiet` tomorrow if you'd like

unixmonkey:
Thank you!

eizengan:
Happy to help. Thanks for maintaining this gem!

unixmonkey:
By the way, this is now in version 1.4.0 of the gem.

eizengan:
Nice! Wasn't expecting to switch my Rails app from my fork so soon - I appreciate the quick turnaround

PR #837, opened 2019-05-30
Fix Webpacker helpers in production env

dleavitt:
`Rails.configuration.assets` will be undefined if asset pipeline is not being used, which leads to an error why trying to check the value of `assets.compile`.

This is primarily a problem when using the webpacker helpers (b/c no sprockets) and only in non-dev environments (b/c this logic gets short-circuited if there's a webpack dev server running.)

`Rails.configuration.assets` is referred to in one other place but it's a block that's only run in Rails 3 so presumably doesn't need to be updated.


unixmonkey:
Sorry it took awhile to look at this. Thanks!

dleavitt:
no problem, thanks for merging!

PR #840, opened 2019-06-12
Fix assets lookup for precompiled assets

SampsonCrowley:
- use asset_manifest if available
  - gems missing from dev dependencies
  - remove static preset asset path in unit test that breaks tests for anyone without ENV variable set

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -6,6 +6,15 @@ class WickedPdf
   module WickedPdfHelper
     module Assets
       ASSET_URL_REGEX = /url\(['"]?([^'"]+?)['"]?\)/
+      AssetStruct = Struct.new(:asset, :filename) do
+        def source
+          asset.respond_to?(:source) ? source.source : asset
+        end
+
+        def to_s
+          asset.to_s
+        end
+      end
 
       def wicked_pdf_asset_base64(path)
         asset = find_asset(path)
@@ -108,6 +117,11 @@ class WickedPdf
       def find_asset(path)
         if Rails.application.assets.respond_to?(:find_asset)
           Rails.application.assets.find_asset(path, :base_path => Rails.application.root.to_s)
+        elsif Rails.application.respond_to?(:assets_manifest) && Rails.application.assets_manifest.respond_to?(:find_sources)
+          asset = Rails.application.assets_manifest.find_sources(path, :base_path => Rails.application.root.to_s).first
+          if asset
+            AssetStruct.new(asset, "#{Rails.application.assets_manifest.dir}/#{Rails.application.assets_manifest.assets[path]}")
+          end
         else
           Sprockets::Railtie.build_environment(Rails.application).find_asset(path, :base_path => Rails.application.root.to_s)
         end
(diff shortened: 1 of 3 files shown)


wicked_pdf.gemspec
```diff
@@ -35,4 +35,6 @@ DESC
   spec.add_development_dependency 'sqlite3', '~> 1.3.6'
   spec.add_development_dependency 'mocha', '= 1.3'
   spec.add_development_dependency 'test-unit'
+  spec.add_development_dependency 'bootsnap'
+  spec.add_development_dependency 'rdoc'
```
unixmonkey: These are only dependencies on newer versions of Rails, and are included in https://github.com/mileszs/wicked_pdf/blob/master/gemfiles/5.2.gemfile and https://github.com/mileszs/wicked_pdf/blob/master/gemfiles/rails_edge.gemfile

test/unit/wicked_pdf_test.rb
```diff
@@ -1,5 +1,5 @@
 require 'test_helper'
-WickedPdf.config = { :exe_path => ENV['WKHTMLTOPDF_BIN'] || '/usr/local/bin/wkhtmltopdf' }
+WickedPdf.config = { :exe_path => ENV['WKHTMLTOPDF_BIN'] }
```
unixmonkey: Is there a compelling reason to remove this default? It's what I use.

lib/wicked_pdf/wicked_pdf_helper/assets.rb
```diff
@@ -108,6 +111,11 @@ def asset_pathname(source)
       def find_asset(path)
         if Rails.application.assets.respond_to?(:find_asset)
           Rails.application.assets.find_asset(path, :base_path => Rails.application.root.to_s)
+        elsif Rails.application.respond_to?(:assets_manifest) && Rails.application.assets_manifest.respond_to?(:find_sources)
+          asset = Rails.application.assets_manifest.find_sources(path, :base_path => Rails.application.root.to_s).first
+          if asset
+            AssetWrapper.new(asset, "#{Rails.application.assets_manifest.dir}/#{Rails.application.assets_manifest.assets[path]}")
+          end
```
unixmonkey: This seems pretty safe and cool the way it's implemented here, but I'm unfamiliar with how it works. Is this new in Rails 6?

wicked_pdf.gemspec
SampsonCrowley: ahh

test/unit/wicked_pdf_test.rb
SampsonCrowley: having a default at all breaks path resolution for non-standard locations. there should be zero need for a hard coded value if the binary is actually installed properly, and if they need a specific location, THEN they can provide an env variable

lib/wicked_pdf/wicked_pdf_helper/assets.rb
SampsonCrowley: Rails 5.2 is when this came into play

test/unit/wicked_pdf_test.rb
SampsonCrowley: ^ At least that's the only thing I can vaguely remember at this point lol I've been using this patch in my own codebase since this pull request, and i can't imagine I'd ever remove a useful default unless it was breaking something else.

FiloSpaTeam:
Even with this fix i have the same problem: 404 error for precompiled assets but, manifest contains them, precompiled asset is deployed and Wicked works outside Mailer.

PR #861, opened 2019-10-30
False-positive

pedrofurtado:
@unixmonkey @mileszs With the option: `raise_on_all_errors: true`, even if pdf is generated successfully, is throwed an exception, because `err` contains the progress of process, with "Done" in the end 👍

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -77,7 +77,7 @@ class WickedPdf
     generated_pdf_file.rewind
     generated_pdf_file.binmode
     pdf = generated_pdf_file.read
-    raise "Error generating PDF\n Command Error: #{err}" if options[:raise_on_all_errors] && !err.empty?
+    raise "Error generating PDF\n Command Error: #{err}" if options[:raise_on_all_errors] && !err.empty? && !err.strip.ends_with?('Done')
     raise "PDF could not be generated!\n Command Error: #{err}" if pdf && pdf.rstrip.empty?
     pdf
   rescue StandardError => e


pedrofurtado:
@marciojg @willradi

unixmonkey:
@pedrofurtado Interesting. It seems that you are invoking also with `quiet: false`, is that right?

Feel free to update the RuboCop config threshold to allow the Travis-CI builds to pass.

pedrofurtado:
@unixmonkey Thanks for feedback! My code is the following:

```ruby
render raise_on_all_errors: true, page_size: 'A4', orientation: 'Portrait', pdf: "my_file", encoding: 'UTF-8', template: "path/to/the/template"
```

pedrofurtado:
@unixmonkey Was it possible to simulate the error? The false-positive :handshake:

pedrofurtado:
@unixmonkey Any news? :+1: :1st_place_medal:

pedrofurtado:
@unixmonkey Was it possible to simulate the error? The false-positive handshake

unixmonkey:
`wkhtmltopdf` by default writes it's output to `stderr`, even when there is no error.

This is stated as by design here (because it can output the PDF on stdout): https://github.com/wkhtmltopdf/wkhtmltopdf/issues/1980

You can change this by specifying the `log_level: 'error'` option, which was introduced here: https://github.com/mileszs/wicked_pdf/pull/834

This should change it so that it only outputs to stderr on actual errors. I wonder if this should be the default since we always output a tempfile in this project.

Can you please check against that combination and see if it helps?

pedrofurtado:
@unixmonkey Thanks for sharing this issue in wkhtmltopdf :handshake: 

The workaround suggested by you works well. The following code (with `log_level: error` together with the rest of parameters does not have false-positive):

```ruby
render log_level: 'error', raise_on_all_errors: true, page_size: 'A4', orientation: 'Portrait', pdf: "my_file", encoding: 'UTF-8', template: "path/to/the/template"
```

But another users that will get the gem inside projects, will get in trouble with this false-positive, if they forget the use of `log_level: error`. Is it valid to merge this PR or provide some workaround inside gem, without the dev need to insert `log_level: error`?

PR #864, opened 2019-11-15
Replace open-uri with net/http

### LINKED ISSUE #863: open-uri is a security issue
OpenURI is a security issue, please avoid using it
https://sakurity.com/blog/2015/02/28/openuri.html

It is currently used in https://github.com/mileszs/wicked_pdf/blob/master/lib/wicked_pdf/wicked_pdf_helper/assets.rb

As it monkey-patches `Kernel.open` it opens up security vulnerabilities in the entire app, not just in the gem itself

shaicoleman:
open-uri is a security vulnerability https://sakurity.com/blog/2015/02/28/openuri.html
Net::HTTP.get is available on all supported rubies >= 2.2

Fixes #863

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -1,4 +1,4 @@
-require 'open-uri'
+require 'net/http'
 # If webpacker is used, need to check for version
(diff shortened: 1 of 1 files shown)


unixmonkey:
This is now in version 2.0.1.

Thank you so much for putting this together!

PR #869, opened 2019-11-26
Support for rails 6

Vasfed:
In rails 6 wicked_pdf initializer causes 
> DEPRECATION WARNING: Initialization autoloaded the constants ActionText::ContentHelper and ActionText::TagHelper.
>
> Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.
> ...

because of early access to `ActionController::Base` and can be fixed by lazy loading.

This PR fixes it by using `ActiveSupport.on_load` (supported since rails 3) 

Also made tests pass on ruby 2.6 and fresh gem versions (had to lock sprockets to 3 because of changed newline handling)

### DIFF
--- a/lib/wicked_pdf/railtie.rb
+++ b/lib/wicked_pdf/railtie.rb
@@ -6,13 +6,15 @@ class WickedPdf
   if defined?(Rails.env)
     class WickedRailtie < Rails::Railtie
       initializer 'wicked_pdf.register', :after => 'remotipart.controller_helper' do |_app|
-        if ActionController::Base.respond_to?(:prepend) &&
-           Object.method(:new).respond_to?(:super_method)
-          ActionController::Base.send :prepend, PdfHelper
-        else
-          ActionController::Base.send :include, PdfHelper
+        ActiveSupport.on_load(:action_controller) do
+          if ActionController::Base.respond_to?(:prepend) &&
+             Object.method(:new).respond_to?(:super_method)
+            ActionController::Base.send :prepend, PdfHelper
+          else
+            ActionController::Base.send :include, PdfHelper
+          end
+          ActionView::Base.send :include, WickedPdfHelper::Assets
         end
-        ActionView::Base.send :include, WickedPdfHelper::Assets
       end
     end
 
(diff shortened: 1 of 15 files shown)


Vasfed:
Fixed CI tests (only rails-edge (6.1 alpha) left because of sprockets' 4 newline, but that's separate story)

kreintjes:
@Vasfed works for me, thanks!

unixmonkey:
@Vasfed Awesome. Thanks for this, and also for tracking down the sprockets newline issue that has been plaguing the test suite.

jdongelmans:
Is there a release planned for this? I'm getting annoyed by the Deprecation Warnings ;)
Thanks!

jdongelmans:
@unixmonkey I've created a PR myself for a new version: https://github.com/mileszs/wicked_pdf/pull/881

unixmonkey:
This is now part of the gem in version 2.0. Thanks so much for the help on this! 🙇

lib/wicked_pdf/railtie.rb
```diff
@@ -6,13 +6,15 @@ class WickedPdf
   if defined?(Rails.env)
     class WickedRailtie < Rails::Railtie
       initializer 'wicked_pdf.register', :after => 'remotipart.controller_helper' do |_app|
-        if ActionController::Base.respond_to?(:prepend) &&
-           Object.method(:new).respond_to?(:super_method)
-          ActionController::Base.send :prepend, PdfHelper
-        else
-          ActionController::Base.send :include, PdfHelper
+        ActiveSupport.on_load(:action_controller) do
```
pierre-alain-b: Could this new approach break the compatibility of wicked_pdf and remotipart? I do observe a bug if wicked_pdf 2.0.x is used with remotipart. In Ruby 2.6, the bug is announced as a FrozenError. We get more information under Ruby <2.5 to see that there is indeed a loop of rendering.

darrenterhune: I can confirm this does break our app. I forked this repo and simply rolled back to a simpler version just to debug this and we are working again https://github.com/darrenterhune/wicked_pdf/commit/ccc861596dcd8e43e5243a79b8658cf7c3c9b4e0 that said, wicked_pdf still worked locally, it was just barfing on CircleCI with this https://github.com/mileszs/wicked_pdf/issues/891

PR #881, opened 2020-01-28
Update to version 2.0.0

jdongelmans:
Updated version to `2.0.0`, so we can have a release which plays nice with Rails 6 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixes
 - None
 
+## [1.5.0] - 2020-01-28
+### New Features
+- [Add rubygems metadata hash to gemspec](https://github.com/mileszs/wicked_pdf/pull/856)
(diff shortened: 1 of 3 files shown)


TejuGupta:
Hi, got to this PR from [here](https://github.com/rails/rails/issues/36546) and learnt that this might fix the ActionText issue we are facing as part of our Rails 6 upgrade. Do you have a tentative date for this release?

unixmonkey:
@jdongelmans After looking this over closely, and integrating into a freshly-generated Rails 6.0.2 app, I think this is good to release. However, because we dropped support for Ruby versions under 2.2, and Rails versions under 4, I think this should be a SEMVER breaking change, and should be version `2.0.0`.

Do you think you could make that change? I'd like you to get some credit for this PR, but don't necessarily want a commit changing to 1.5.0 and that never getting a related release.

jdongelmans:
@unixmonkey you are right, I've changed the version to `2.0.0` and moved the relevant issue to the Breaking changes section in `CHANGELOG.md`

unixmonkey:
Version 2.0 is now released. Thanks for the help and the push to get this out there.

PR #889, opened 2020-02-24
Added option to skip the prefix of temp files with file:///

bo-oz:
Made two changes:

1/ Added option `option[:slip_file_prefix]` that prevents `pdf_from_string` from calling .`pdf_from_html_file` and calls `pdf_from_url` directly.
2/ This option also influences the temp file that is created for header/footer content
3/ Passed the `options[:temp_path]` to the Tempfile that is being created for the header/footer content.


unixmonkey:
@bo-oz Why is this desirable?

PR #896, opened 2020-03-25
Support webpacker assets

davegudge:
Introduces a helper method to access webpacked assets directly.

e.g:

`image_tag wicked_pdf_asset_pack_path("media/images/foobar.png")`


unixmonkey:
Nice! Thanks 🙇

PR #903, opened 2020-04-16
Support for --disable-local-file-access to stop SSRF

fawad:
- When we are depended on taking HTML as an input from the user, they can add malicious content, and fetch any system file by doing something like:
```
"><iframe src="http://SOME_IP/latest/meta-data/public-keys/0/openssh-key" height="500" width="500">
``` 

- This can be fixed by passing ```--disable-local-file-access``` to **wkhtmltopdf**

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ class ThingsController < ApplicationController
                disable_external_links:         true,
                print_media_type:               true,
                disable_smart_shrinking:        true,
+               disable_local_file_access:      true,
                use_xserver:                    true,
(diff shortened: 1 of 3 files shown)


unixmonkey:
This work was also done as part of https://github.com/mileszs/wicked_pdf/pull/920, and no longer needed. Thank you for doing it though, and sorry for not getting to it earlier. The other PR got my attention earlier because the default for this option has changed in `wkhtmltopdf` 0.12.6 to be disabled by default.

PR #909, opened 2020-04-30
Add option `use_xvfb` to emulate an X server

mschnitzer:
Hi,

so I really struggled to get `wkhtmltopdf` (which is obviously used by `wicked_pdf`) working without having an X server running on my virtual server. Unfortunately, the best solution I found, was to wrap all `wkhtmltopdf` around the `xvfb-run` command which is - from my understanding - emulating an X server for systems where a real X server is not available. Doing so, made my PDF's generating normally again.

I'm not really sure whether there's any better solution. I didn't find anything working so far besides the approach I'm following.

As there are many topics/issues about generating PDF's without an X server being available, I decided to propose my code changes to `wicked_pdf`. Let's discuss whether this is a good approach or if you see anything else, which seems to be more suitable.

____
Commit Message:

On some systems an X server is not available, and thus a PDF generation
is not possible. This can be avoided when installing `xvfb-run` which
emulates an X server.

Enabling the `use_xvfb` option will wrap any `wkhtmltopdf` commands
around the command `xvfb-run`, in order to be able to create PDF's, even
without having an X server running.

### DIFF
--- a/generators/wicked_pdf/templates/wicked_pdf.rb
+++ b/generators/wicked_pdf/templates/wicked_pdf.rb
@@ -18,4 +18,10 @@ WickedPdf.config = {
   # Layout file to be used for all PDFs
   # (but can be overridden in `render :pdf` calls)
   # layout: 'pdf.html',
+
+  # Using wkhtmltopdf without an X server can be achieved by enabling the
+  # 'use_xvfb' flag. This will wrap all wkhtmltopdf commands around the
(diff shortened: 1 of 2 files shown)


mathieujobin:
LGTM, rubocop failures looks unrelated

unixmonkey:
This is now in the published gem version 2.1.0.

Thank you for your help!

PR #916, opened 2020-05-25
Prevent double inclusion of modules

georf:
In my project the `on_load` was called multiple times. This ended with fatals on loading views. I prevent to included the modules multiple times.


georf:
Shit. This was not the entire problem.

PR #918, opened 2020-05-28
Allow cover page to have header/footer

Hkly:
Allows cover section to be either cover page object or page page object.
If is page page object, page will include header/footer.


Hkly:
Didn't mean to open this! Having trouble running the tests locally.

Issue #917, opened 2020-05-28
Incompatible with view_component gem


woller:
## Issue description
In a project using both [view_component](https://github.com/github/view_component) and wicked_pdf, trying to render a pdf throws a fatal error. I suspect the issue is related to this other issue: https://github.com/mileszs/wicked_pdf/issues/891

## System specifications

wicked_pdf gem version (output of `cat Gemfile.lock | grep wicked_pdf`): 2.0.2

wkhtmltopdf version (output of `wkhtmltopdf --version`): 12.5

whtmltopdf [provider gem](https://rubygems.org/search?utf8=%E2%9C%93&query=wkhtmltopdf) and version if one is used: 0.12.5.4

platform/distribution and version (e.g. Windows 10 / Ubuntu 16.04 / Heroku cedar): MacOS Catalina

## Stacktrace
```ruby
fatal (exception reentered):
  
#<Thread:0x00007f8c36739348@puma threadpool 002 /Users/woller/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/puma-4.3.5/lib/puma/thread_pool.rb:89 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
/Users/woller/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/web-console-4.0.2/lib/web_console/middleware.rb:17:in `catch': exception reentered (fatal)
/Users/woller/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/web-console-4.0.2/lib/web_console/middleware.rb:17:in `catch': can't modify frozen fatal: #<fatal: exception reentered> (FrozenError)
	87: from /Users/woller/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/puma-4.3.5/lib/puma/thread_pool.rb:134:in `block in spawn_thread'
	86: from /Users/woller/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/puma-4.3.5/lib/puma/server.rb:328:in `block in run'
	85: from /Users/woller/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/puma-4.3.5/lib/puma/server.rb:472:in `process_client'
	84: from /Users/woller/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/puma-4.3.5/lib/puma/server.rb:713:in `handle_request'
[...]

PR #919, opened 2020-06-11
Fix issue with render method in PdfHelper

Okatawa:
Fix https://github.com/mileszs/wicked_pdf/issues/917


Okatawa:
Pull request fix it https://github.com/mileszs/wicked_pdf/issues/917

yhirano55:
@mileszs Could you check this patch...? :pray:

PR #920, opened 2020-06-11
Add a option to disable/enable local file access, that is enabled by default in 0.12.5 and false by default in 0.12.6 version of wkhtmltopdf

pedrofurtado:
@unixmonkey  @mileszs  👍 

The 0.12.6 version of wkhtmltopdf  has a breaking change option, that is true by default, related to local file access. 

Link for breaking change in WkHtmlToPDF 0.12.6:
https://github.com/wkhtmltopdf/wkhtmltopdf/releases/tag/0.12.6
https://github.com/wkhtmltopdf/wkhtmltopdf/issues/4536
https://github.com/wkhtmltopdf/wkhtmltopdf/blob/0.12.6/src/shared/commonarguments.cc#L227-L228

The local file access, that was enabled by default in <= 0.12.5, now is false by default in 0.12.6+

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,8 +188,8 @@ class ThingsController < ApplicationController
                print_media_type:               true,
 
                # define as true the key 'disable_local_file_access' or 'enable_local_file_access', not both
-               disable_local_file_access:      false,
-               enable_local_file_access:       true,
+               disable_local_file_access:      true,
+               enable_local_file_access:       false,
 
                disable_smart_shrinking:        true,
                use_xserver:                    true,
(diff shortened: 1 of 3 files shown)


pedrofurtado:
Here, we tested in our applications, and it worked well 🎉

unixmonkey:
This is now in the published gem version 2.1.0.

Thanks again for your help!

Kriechi:
Whats the proposed solution for `wicked_pdf_image_tag` to render a local file, e.g., page header graphics?

PR #925, opened 2020-06-29
Remove code for unsupported rails and ruby versions

### LINKED ISSUE #111: wicked_pdf and remotipart gems can't coexist

fsateler:
Since rails < 4 and ruby < 2.2 are unsupported, we can assume that `ActionController::Base.after_action` and `Module.prepend` exist.

This fixes https://github.com/github/view_component/issues/288 , and should fix #111 (but didn't test)

### DIFF
--- a/lib/wicked_pdf/pdf_helper.rb
+++ b/lib/wicked_pdf/pdf_helper.rb
@@ -8,11 +8,7 @@ class WickedPdf
       base.class_eval do
         alias_method_chain :render, :wicked_pdf
         alias_method_chain :render_to_string, :wicked_pdf
-        if respond_to?(:after_action)
-          after_action :clean_temp_files
-        else
-          after_filter :clean_temp_files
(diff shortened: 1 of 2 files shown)


unixmonkey:
Thank you very much for this. I'm prepping for a Semver major release around dropping old Ruby & Rails versions, so it's time for this now. Sorry it took so long.

Issue #288, opened 2014-02-19
Large pdf generation header and footer fails


ushavimal:
I am using wicked_pdf in my app. I am able to generate small pdfs with header and footer without any problem. But when I try to generate a large pdf, it fails with header and footer. Was able to generate large files without header and footer.

Below is my options to generate pdf

    pdf = render_to_string :pdf => pdf_request.file_name,
           :layout => "pdf.html",
           :template => "shared/report",
           :wkhtmltopdf => "/usr/bin/wkhtmltopdf",
           :formats => [:pdf],
           :margin => {:top => 30},
           :header => {:html => {:template => 'shared/pdf_header',
                               :locals => {:title => pdf_request.title, :date => date}},
                               :spacing => 10},
           :footer => {:html => {:template => 'shared/pdf_footer',
                            :locals => {:run_at => TimeUtilities.sanitize_display_time(Time.now).to_s(:human)}},
                            :spacing => -4,
                            :line => true}

And the error I am getting is

    "***************\"/usr/bin/wkhtmltopdf\" -q  --header-spacing 10 --header-html \"file:////tmp/wicked_header_pdf20140219-14233-nb2cxa.html\" --footer-spacing -4 --footer-line  --footer-html \"file:////tmp/wicked_footer_pdf20140219-14233-11395q7.html\"    --margin-top 30    \"file:////tmp/wicked_pdf20140219-14233-128ah5k.html\" \"/tmp/wicked_pdf_generated_file20140219-14233-1g0xgbu.pdf\" ***************"
    2014-02-19T23:16:43Z 14233 TID-1ekr1w PdfWorker JID-34a78c3af35d06ff7ced9506 ERROR: RuntimeError: Failed to execute:
    "/usr/bin/wkhtmltopdf" -q  --header-spacing 10 --header-html "file:////tmp/wicked_header_pdf20140219-14233-nb2cxa.html" --footer-spacing -4 --footer-line  --footer-html "file:////tmp/wicked_footer_pdf20140219-14233-11395q7.html"    --margin-top 30    "file:////tmp/wicked_pdf20140219-14233-128ah5k.html" "/tmp/wicked_pdf_generated_file20140219-14233-1g0xgbu.pdf" 
    Error: PDF could not be generated!
[...]

PR #947, opened 2020-10-21
Add `wicked_pdf_url_base64`.

joshuapinter:
_We're using this in production to solve our problem. Looking to get some feedback from @mileszs before finalizing this properly and, hopefully, getting it merged to help others in similar situations._

Using URLs to reference images can cause a lot of problems with wkhtmltopdf, particularly when used in the header and footer. It produces errors such as "Too many open files" as well as buffer/stack overflows. These issues can be seen here, among other places: 

- https://github.com/mileszs/wicked_pdf/issues/288
- https://github.com/wkhtmltopdf/wkhtmltopdf/issues/1819
- https://github.com/wkhtmltopdf/wkhtmltopdf/issues/3019
- https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2093
- https://github.com/wkhtmltopdf/wkhtmltopdf/issues/4762
- https://github.com/wkhtmltopdf/wkhtmltopdf/issues/3081
- https://github.com/wkhtmltopdf/wkhtmltopdf/issues/4296

As described [here](https://github.com/mileszs/wicked_pdf/issues/288#issuecomment-130106916), one of the key solutions to avoiding this issue is providing the image encoded as base64. This helper method takes a URL of an image, opens the URL, reads the image data and encodes it to base64.

If the URL cannot be open (`OpenURI::HTTPError` is raised), it will log a warning and return `nil`.

### TODOs:

- [ ] Better placement (since this is not technically part of the asset pipeline)?
- [ ] Better naming?
- [ ] Tests.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -15,6 +15,20 @@ class WickedPdf
         "data:#{asset.content_type};base64,#{Rack::Utils.escape(base64)}"
       end
 
+      # Using `image_tag` with URLs when generating PDFs (specifically large PDFs with lots of pages) can cause buffer/stack overflows.
+      #
+      def wicked_pdf_url_base64(url)
+        response = Net::HTTP.get_response(URI(url))
+
+        if response.is_a?(Net::HTTPSuccess)
+          base64 = Base64.encode64(response.body).gsub(/\s+/, '')
+          "data:#{response.content_type};base64,#{Rack::Utils.escape(base64)}"
+        else
+          Rails.logger.warn("[wicked_pdf] #{response.code} #{response.message}: #{url}")
+          nil
+        end
+      end
+
       def wicked_pdf_stylesheet_link_tag(*sources)
         stylesheet_contents = sources.collect do |source|
           source = WickedPdfHelper.add_extension(source, 'css')


lib/wicked_pdf/wicked_pdf_helper/assets.rb
```diff
@@ -15,6 +15,17 @@ def wicked_pdf_asset_base64(path)
         "data:#{asset.content_type};base64,#{Rack::Utils.escape(base64)}"
       end
 
+      # Using `image_tag` with URLs when generating PDFs (specifically large PDFs with lots of pages) can cause buffer/stack overflows.
+      #
+      def wicked_pdf_url_base64(url)
+        asset = URI.parse(url).open
```
unixmonkey, line 21: Does `read_from_uri` work here instead? [This project has been cautioned against using URI#open](https://github.com/mileszs/wicked_pdf/issues/863) for security reasons.

joshuapinter: Let me try it out. I original just had `open()` and Rubocop cautioned against it and in favour of `URI.parse`. Hang on...

joshuapinter: Okay, using `read_from_uri` won't work because it just returns a `String` and we need the proper response in order to get the `content_type`.

I was able to get it to work using: 

```ruby
response = Net::HTTP.get_response(URI(url))
base64 = Base64.encode64(response.body).gsub(/\s+/, '')
"data:#{response.content_type};base64,#{Rack::Utils.escape(base64)}"
```

Does that work?

_P.s. We're still on Rails 4.2 but this made me open up an Issue with our own codebase to replace all uses of `URI.parse`._

joshuapinter: Okay, updated the code and also had to change how failed responses were handled.

Note, I wanted to setup a guard like: 

```ruby
if !response.is_a?(Net::HTTPSuccess)
  Rails.logger.warn("[wicked_pdf] #{response.code} #{response.message}: #{url}")
  return nil
end
```

But your Rubocop config wanted me to change it to: 

```ruby
unless response.is_a?(Net::HTTPSuccess)
  Rails.logger.warn("[wicked_pdf] #{response.code} #{response.message}: #{url}")
  nil
end
```

And I despise that with a passion (we have that Cop changed in our Rubocop config) so went with the `if/else` instead as it's much more clear.

unixmonkey (approved):
Thanks for this!

joshuapinter:
:+1: Thanks for a great library.

Issue #961, opened 2021-01-26
wicked_pdf_javascript_pack_tag appears to insert a generic html example document


mfrederickson:
## Issue description

It appears to be inserting an "example domain" html document instead of my javascript file.

When my layout contained `stylesheet_pack_tag` and `javascript_pack_tag` it produced the following:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Surveyor</title>
    <!-- stylesheet_pack_tag -->
    
    <!-- javascript_pack_tag -->
    <script src="/packs/js/application-4f9ed7c5710b7f159806.js"></script>
  </head>
  <body>
    <main role="main" class="container-fluid application-">
...
```

But when I added the `wicked_pdf_` prefix to those, it produced:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Surveyor</title>
    <!-- wicked_pdf_stylesheet_pack_tag -->
    <style type='text/css'></style>
   
    <!-- wicked_pdf_javascript_pack_tag -->
    <script type='text/javascript'><!doctype html>
<html>
<head>
    <title>Example Domain</title>

    <meta charset="utf-8" />
    <meta http-equiv="Content-type" content="text/html; charset=utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <style type="text/css">
    body {
        background-color: #f0f0f2;
        margin: 0;
        padding: 0;
        font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
        
    }
    div {
        width: 600px;
        margin: 5em auto;
        padding: 2em;
        background-color: #fdfdff;
        border-radius: 0.5em;
        box-shadow: 2px 3px 7px 2px rgba(0,0,0,0.02);
    }
    a:link, a:visited {
        color: #38488f;
        text-decoration: none;
    }
    @media (max-width: 700px) {
        div {
            margin: 0 auto;
            width: auto;
        }
    }
    </style>    
</head>

<body>
<div>
    <h1>Example Domain</h1>
    <p>This domain is for use in illustrative examples in documents. You may use this
    domain in literature without prior coordination or asking for permission.</p>
    <p><a href="https://www.iana.org/domains/example">More information...</a></p>
</div>
</body>
[...]

Issue #923, opened 2020-06-21
Using WickedPDF with Webpacker & Bulma CSS


DaniG2k:
## Issue description

I'm trying to use WickedPDF in a Rails 6 project that uses the [Bulma CSS](https://bulma.io/) framework. However, I cannot get it to work properly and I'm wondering what I am doing wrong.

I have a `pdf.html.slim` layout file with the following:

```slim
doctype html
html
  head
    meta charset="utf-8"
    meta name="viewport" content="width=device-width, initial-scale=1"
    title= content_for?(:title) ? content_for(:title) : t('website.title')
    = wicked_pdf_stylesheet_pack_tag 'pdf'
    = wicked_pdf_javascript_pack_tag 'pdf'

  body
    .containers
      .container
        = yield
```

My `app/javascript/packs/pdf.js` file looks like this:

```javascript
import 'stylesheets/pdf';
```

which in turn calls `app/javascript/stylesheets/pdf.css`

```scss
@import 'colors';

// ~ to tell webpack that this is not a relative
// import (but in node_modules dir):
@import '~bulma';
```

When I open a PDF, however, none of the Bulma styling is applied.

I'm using a vanilla `config/webpacker.yml` file:

```yml
# Note: You must restart bin/webpack-dev-server for changes to take effect

default: &default
  source_path: app/javascript
  source_entry_path: packs
  public_root_path: public
  public_output_path: packs
  cache_path: tmp/cache/webpacker
  check_yarn_integrity: false
  webpack_compile_output: true

  # Additional paths webpack should lookup modules
  # ['app/assets', 'engine/foo/app/assets']
  resolved_paths: []

  # Reload manifest.json on all requests so we reload latest compiled packs
  cache_manifest: false

  # Extract and emit a css file
  extract_css: false

  static_assets_extensions:
    - .jpg
    - .jpeg
    - .png
    - .gif
    - .tiff
    - .ico
    - .svg
    - .eot
    - .otf
    - .ttf
    - .woff
    - .woff2

  extensions:
    - .vue
    - .mjs
    - .js
    - .sass
    - .scss
    - .css
    - .module.sass
    - .module.scss
    - .module.css
    - .png
    - .svg
    - .gif
    - .jpeg
    - .jpg

development:
  <<: *default
  compile: true

  # Verifies that correct packages and versions are installed by inspecting package.json, yarn.lock, and node_modules
  check_yarn_integrity: true

[...]

Issue #860, opened 2019-10-22
Webpack helpers does not work


EdouardPan:
## Issue description

In wicked_pdf doc, it is said that I can use ``wicked_pdf_stylesheet_pack_tag`` and ``wicked_pdf_javascript_pack_tag`` to include my stylesheets and javascript from webpack but nothing works.

In a rails project with webpack, here is the code from the controller:
```ruby
      format.pdf do
        render template: "pdf_reports/show", 
        layout: "wicked_layout",
        pdf: "report"
      end
```

Here is the code from the layout:
```ruby
<!DOCTYPE html>
    <html>
        <head>
           <%= csrf_meta_tags %>
           <%= wicked_pdf_javascript_pack_tag 'application', 'data-turbolinks-track': 'reload' %>
           <%= wicked_pdf_stylesheet_pack_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
           <%= wicked_pdf_stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
      </head>
      <body>
      	<%= yield %>
      </body>
</html>
```

Here is the code from the view pdf.erb:
```html
<h1 class="text-red-base">Test pdf</h1>
<h2 class="test-wicked">mldgmdjgfd</h2>
```

## Expected or desired behavior

It works with wicked_pdf_stylesheet_link_tag (test-wicked is applied from sprockets: text is blue) but not with wicked_pdf_stylesheet_pack_tag (h1 should be red but is not). 

PS: I posted originally the question on stackoverflow but I think it may more sens to address an issue.

## System specifications

wicked_pdf gem version (output of `cat Gemfile.lock | grep wicked_pdf`): 1.4.0

wkhtmltopdf version (output of `wkhtmltopdf --version`): 0.12.4

whtmltopdf [provider gem](https://rubygems.org/search?utf8=%E2%9C%93&query=wkhtmltopdf) and version if one is used:
wkhtmltopdf-binary (0.12.4)

platform/distribution and version (e.g. Windows 10 / Ubuntu 16.04 / Heroku cedar):
Ubuntu 18.04

Thank you in advance!

unixmonkey:
Cross-referencing your StackOverflow post here:
https://stackoverflow.com/questions/58490299/how-to-include-css-stylesheet-into-wicked-pdf

There's no answer there. Did you ever figure this out or find a workaround?

EdouardPan:
[...]

PR #973, opened 2021-04-23
fix url generating in webpacker helper

vkalach:
There are a lot of issues that are connected with this fix.
Some of them: https://github.com/mileszs/wicked_pdf/issues/961, https://github.com/mileszs/wicked_pdf/issues/923, https://github.com/mileszs/wicked_pdf/issues/860

Most of the projects don't use the `asset_host` config. I think it can help many people to avoid monkey patching and so on.


unixmonkey (approved):
Great! Thank you for this.

PR #974, opened 2021-04-24
Readme - added more useful readings

yshmarov:
info that I personally found really useful when learning how to use the gem + my experience


unixmonkey:
Thank you!

PR #976, opened 2021-05-06
readme - add links to couple of screencasts

yshmarov:
I wanted to also add driftingruby (their wicked_pdf video is under a paywall) & gorails (they don't have)


unixmonkey (approved):
Thank you!

PR #986, opened 2021-08-24
GitHub actions

pedrofurtado:
Hi, @unixmonkey ! 👋 

I noticed that CI is red/failing for latest commits in repo. I thought that can be a good idea to switch to Github Actions CI, in order to make it green again and also not depends on Travis CI (that changed policies/usage limits recently).

For a while, I've added to CI only stable rails versions, and versions that are maintained at this moment.

Let me know what do you think about 🤝 🍻  Your eyes/review on it will be awesome 🕺 

Thanks a lot!

PR #987, opened 2021-08-31
Finish GitHub actions transition

pedrofurtado:
Hello, @unixmonkey ! 👋 

This PR finishes the Github Actions transition started at #986 🎉 

New cops are present in recent rubocops versions (that is needed for Ruby 3 support), but initially them are disabled. Later, it will be possible to "turn on" each one of them and change the source code to be compliant with each rule. But is better to make it gradually 🤝 

Let me know your thoughts about it 🕺 

Thanks! 🤓


wicked_pdf.gemspec
```diff
@@ -21,7 +21,7 @@ DESC
     'changelog_uri' => 'https://github.com/mileszs/wicked_pdf/blob/master/CHANGELOG.md'
   }
 
-  spec.required_ruby_version = Gem::Requirement.new('>= 2.2')
+  spec.required_ruby_version = Gem::Requirement.new('>= 2.6')
```
unixmonkey, line 24: Is there a reason to raise this minimum Ruby version, and by so much? I'd like people on older rubies to still be able to use this gem.

pedrofurtado: We can maintain this old ruby version, if you prefer, no problem 🤝 But CI tests may not guarantee it works well on such older versions of the gem

pedrofurtado: I made rollback on this change, returning to ruby 2.2 👍

mathieujobin: My two cents here, I usually keep CI to run for old versions as long as tests pass.
Once we hit a failure, then we take the decision if it's worth fixing or moving on, deprecating the failing version.

gemfiles/rails_edge.gemfile
```diff
@@ -1,9 +0,0 @@
-source 'https://rubygems.org'
-
-gem 'bootsnap' # required to run `rake test` in Rails 6.0
-gem 'mocha', '= 1.3' # newer versions blow up
-gem 'rails', git: 'https://github.com/rails/rails.git'
-gem 'rdoc'
-gem 'sqlite3', '~> 1.4'
-
-gemspec path: '../'
```
mathieujobin, line 9: Why not keeping Rails Edge? 7.0 is just around the corner. I have an app running it already.
Maybe just allow-failures on edge

PR #992, opened 2021-11-09
Include view helper on view load.

emielvanlankveld:
We ran into some problems using this gem with our Rails 6 app that adds a lot of custom paths to Zeitwerk's autoloaders. Narrowed it down to the view helper being included in `ActionView::Base` during the action_controller load hook rather than in the action_view load hook.

This change fixed the issue for us, and it feels better than accessing a private method.

### DIFF
--- a/lib/wicked_pdf/railtie.rb
+++ b/lib/wicked_pdf/railtie.rb
@@ -6,10 +6,8 @@ class WickedPdf
   if defined?(Rails.env)
     class WickedRailtie < Rails::Railtie
       initializer 'wicked_pdf.register', :after => 'remotipart.controller_helper' do |_app|
-        ActiveSupport.on_load(:action_controller) do
-          ActionController::Base.send :prepend, PdfHelper
-          ActionView::Base.send :include, WickedPdfHelper::Assets
-        end
+        ActiveSupport.on_load(:action_controller) { ActionController::Base.send :prepend, PdfHelper }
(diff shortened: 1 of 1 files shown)


unixmonkey:
Thanks!

mathieujobin:
lots of good fixes on `master`, is a new release imminent ?

unixmonkey:
@mathieujobin Yes. I am planning a release in the next few days. If you’d like to help, I still need to comb through the changes and add them to the CHANGELOG, and determine if any are breaking. Ideally also making it work (or verifying it works) with Rails 7 and esbuild assets.

mathieujobin:
There has been several PR merges since the last release 18 months ago... quite a bit to review in one go...

mathieujobin:
@unixmonkey maybe you can release current master as 3.0.0
and if we need to make any more 2.2+ later, we can do that... otherwise, that can be considered a new major

ruby 1.x and rails 2.x has been deprecated, all the way until rails 7 support ;)

https://github.com/mileszs/wicked_pdf/compare/2.1.0...master
https://rubygems.org/gems/wicked_pdf
https://github.com/mileszs/wicked_pdf/tags

mathieujobin:
if you feel this isn't ready for 3.0...
you can bump to 2.5 leaving some gap before and after

or 2.90, KDE style ;)

leonelgalan:
A similar issue was reported in #938 (closed), I thought linking the two might be a good idea jut in case someone experiences the issues described and finds that one first.

Issue #891, opened 2020-03-09
Thread puma threadpool terminated with exception machine stack overflow in critical region (fatal)


darrenterhune:
## Issue description

Our CI server is having a problem with the recent major bump to 2.0. Locally we are fine and pdf's work. If I downgrade to 1.4.0 our CI is fine.

Seems this was also an issue in the past: https://stackoverflow.com/questions/55280153/machine-stack-overflow-in-critical-region-fatal
https://github.com/mileszs/wicked_pdf/issues/810

```
Randomized with seed 34635
Capybara starting Puma...
* Version 4.3.3 , codename: Mysterious Traveller
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:34469
#<Thread:0x0000561dc0601600@puma threadpool 001@/home/circleci/project/vendor/bundle/ruby/2.6.0/gems/puma-4.3.3/lib/puma/thread_pool.rb:89 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
/home/circleci/project/vendor/bundle/ruby/2.6.0/gems/rack-2.2.2/lib/rack/urlmap.rb:58:in `each': machine stack overflow in critical region (fatal)
```

https://github.com/puma/puma/blob/master/lib/puma/thread_pool.rb#L89

## Expected or desired behavior

CI runs and passes

## System specifications

wicked_pdf 2.0.1
wkhtmltopdf-binary-edge 0.12.4.0 because of https://github.com/pallymore/wkhtmltopdf-binary-edge/issues/15 and extra build config just to get it to work so we locked it
Ubuntu 16.04
ruby 2.6.5
rails 6.0.2.1

unixmonkey:
The actual code changes between 1.4.0 and 2.0.1 are pretty small, and I suspect this issue may have to do with a conflict with some other gem that hooks into `render`, like `remotipart` (do you have this in your project?).

Please read [this comment](https://github.com/mileszs/wicked_pdf/issues/111#issuecomment-592965281) and look at the diff to see what I'm talking about. I'm curious why this affects CI, but not development. Is CI running the same version of Ruby you develop with? Does it precompile assets, where maybe you don't locally?

darrenterhune:
Interesting, yeah I didn't see that comment. Let me dive back into this and I'll get back to you. Thanks for some direction 👍

darrenterhune:
[...]

PR #993, opened 2021-11-18
fix: `webpacker_source_url`

davegudge:
Since the change in #973, I've experienced an issue where images are not
being embedding in the pdf generated (in non development environments)
when using the following:

```
image_tag wicked_pdf_asset_pack_path("static/example.png")
```

For context: The pdfs are being generated via a background job on a
separate sever to the webapp, so using the `asset_host` option is not
ideal, as the pdf would then need to fetch the assets from another server
when they're already available locally. Although not ideal, I have tried setting
`Rails.application.config.asset_host` via `production.rb` to no avail.

This commit reinstates the original functionality prior to #973, whist
still retaining the new functionality add via #973.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -191,7 +191,11 @@ class WickedPdf
 
         # In Webpacker 3.2.0 asset_pack_url is introduced
         if Webpacker::VERSION >= '3.2.0'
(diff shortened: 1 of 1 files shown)


unixmonkey (approved):

davegudge:
@unixmonkey Thanks for actioning so quickly. I think there would be merit in releasing a new version (if possible)? If only to close #891, as I think there are a few people tracking `master`.


I was asked on https://github.com/mileszs/wicked_pdf/issues/891#issuecomment-933129168 if _I_ could release a new version a while back 😄

PR #998, opened 2021-12-19
CI tweaks

unixmonkey:


mathieujobin:
@unixmonkey You have added the rails 7 file in your first commit, then you removed it in your second commit ...

PR #1007, opened 2022-01-26
Try to fetch asset on filesystem first before potentially going over the network

Mathiou04:
## Context

Currently for some rails configuration, the `read_asset` method (used in methods like `wicked_pdf_stylesheet_link_tag`, allowing to inline CSS/JS inside pdf template) goes over the network to fetch assets.

For example, when assets are precompiled and Rails is configured to use a specific asset host (a CDN for example), the `read_asset` method retrieves the file through an HTTP call to the CDN.

This may be a waste of resource in cases when the asset is still available on the filesystem.
Moreover, the underlying HTTP library used (`Net::HTTP`) is not raising any exception if the response code is > 299.
If for any reason you get a 3XX (the library is not following redirects either), 4XX, 5XX, the css/js included in the pdf may be empty/invalid. The generated pdf will be (silently!) completely messed up.

## Intent 
This PR attempts at trying to improve this by first trying to retrieve the asset from the filesystem (generally the pre-compiled asset in the public folder) and read it if available, otherwise falling back on the usual mechanism if the asset is not found.

I am far from clearly aware of the consequences of this change, but wanted to illustrate my reasoning. 

Any advice/feedback welcome! 

Also, fixing the issue with the silent errors might also be something worth considering.


Mathiou04:
Hi @mathieujobin 
Thanks! This is done, waiting for approval.

mathieujobin:
after I removed the trailing space, and merged with latest `master`

all tests passed https://github.com/payrollhero/wicked_pdf/actions/runs/2372354673

d4rky-pl:
@mileszs any plans to integrate this?

mathieujobin (approved):
lets merge this and cut a release

PR #1010, opened 2022-02-18
Add support for Propshaft asset pipeline

serggl:
## Purpose

This adds support to [rails/propshaft](https://github.com/rails/propshaft) asset pipeline introduced recently.

## Solution

Basically it wraps `Propshaft::Assets` class and adds some methods to be compatible with existing `Sprockets::Asset` api


tmaier:
I tried it. I can confirm it works.

unixmonkey (approved):
Thank you!

Issue #1004, opened 2022-01-21
Publish a new release


mathieujobin:
## Issue description
no official release was made in 18 months+

creating an issue to continue discussion started in 
https://github.com/mileszs/wicked_pdf/pull/992

## Expected or desired behavior
We should have one or more release.

mathieujobin:
@mileszs Do you need help making a release?

unixmonkey:
@mathieujobin You can help by updating the CHANGELOG with summaries of the changes that have been merged. I'm happy to put out a release, but I haven't paid enough attention to the big picture in awhile to see if it needs to be a SEMVER breaking change version or not. What I don't want is to release a new patch or feature version that breaks apps.

mathieujobin:
@unixmonkey have you had a chance to look at @1013 ?

Thanks

mathieujobin:
Let me know if you need anything else

unixmonkey:
Released as 2.6.0. Thanks for your help!

PR #1013, opened 2022-03-28
Update changelog with all versions

mathieujobin:
There you go, I went through the whole history. And mark a few points that could have been releases.

I have tagged these that I noted locally. If you want, I can push and build .gem files if that is helpful.

Hopefully solves #1004


mathieujobin:
@unixmonkey had a chance to look at this?

mathieujobin:
THanks a lot for reviewing this and merging this. I'm excited to see what's next

PR #1019, opened 2022-05-02
"require 'delegate'" for compatibility with ruby 2.7

BryanPierce:
When running wicked_pdf with ruby 2.7 in a non-rails environment the following error is encountered:
`uninitialized constant WickedPdf::WickedPdfHelper::Assets::SimpleDelegator`

The delegate library was loaded by default in older versions of ruby, but in 2.7 this is not the case. Explicitly requiring delegate in assets.rb resolves the issue.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -1,4 +1,5 @@
 require 'net/http'
+require 'delegate'
 
 class WickedPdf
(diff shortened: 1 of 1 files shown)


unixmonkey (approved):
Thank you!

unixmonkey:
This is now released in version 2.6.2. Please let me know if it works for you.
Thank you!

BryanPierce:
It works. Thanks!

Issue #1031, opened 2022-06-21
Header/footer temporary file is removed before wkhtmltopdf is called


hpoo:
## Issue description
When creating a pdf-file with content given in footer- and header-options, the pdf creation sometimes fail. 
The reason for the failure is that temporary files widget-pdf creates from the header/footer content are removed before the `wkhtmltopdf` binary is called. `wkhtmltopdf` fails with errors like
`Error: Failed to load file:////tmp/wicked_header_pdf20220620-3625428-872dqb.html?`

Further investigating the issue it seems that the culprit may be that only the temporary file name is stored to options hash in the `options_parser#parse_header_footer`. The temporary file objects are not stored and are loose for the JRuby garbage collection to clean up and remove. 

## Expected or desired behavior
It would be expected that the temporary file live long enough to be available for the `wkhtmltopdf` to use

## System specifications
wicked_pdf gem version:wicked_pdf-2.6.3
wkhtmltopdf version:wkhtmltopdf 0.12.4
platform/distribution and version (e.g. Windows 10 / Ubuntu 16.04 / Heroku cedar):
ubuntu 18.04, Rails 6.0 with JRuby 9.3.3.0

thanhcuong1990:
I got the same issue with:
`wicked_pdf: 2.6.3`
`wkhtmltopdf 0.12.6 (with patched qt)`
`macOS Montery`

The header/ footer html files is deleted before the pdf generated. 
`wicked_header_pdfxxxxx.html`
`wicked_footer_pdfxxxxxx.html`

GalenkoEugene:
I got the same issue with:
`ruby:2.6.6-slim`
`wicked_pdf:  2.1.0 and 2.6.3`

unixmonkey:
The tempfiles are pushed into an [instance variable](https://github.com/mileszs/wicked_pdf/blob/master/lib/wicked_pdf/pdf_helper.rb#L110) that only gets [cleared after the render action](https://github.com/mileszs/wicked_pdf/blob/master/lib/wicked_pdf/pdf_helper.rb#L9). 

How do you think that is happening early?

thanhcuong1990:
@unixmonkey I was render header/footer by `render_to_string` described in [Super Advanced Usage](https://github.com/mileszs/wicked_pdf#super-advanced-usage). Like below:

```ruby
# create a pdf from string using templates, layouts and content option for header or footer
pdf = WickedPdf.new.pdf_from_string(
[...]

PR #1039, opened 2022-10-09
Fix header/footer temp file is removed before wkhtmltopdf called

thanhcuong1990:
Fix issue #1031. 

Header/footer generated by `WickedPdf::Tempfile` is deleted before wkhtmltopdf called. 
This PR try to use normal `File` then delete it when the pdf file generated. 

Temporary fix in your Gemfile.
`gem 'wicked_pdf', git: 'https://github.com/thanhcuong1990/wicked_pdf.git', branch: 'master'`

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -46,13 +46,14 @@ class WickedPdf
     string_file.close if string_file
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
   def pdf_from_url(url, options = {})
     # merge in global config options
     options.merge!(WickedPdf.config) { |_key, option, _config| option }
     generated_pdf_file = WickedPdf::Tempfile.new('wicked_pdf_generated_file.pdf', options[:temp_path])
     command = [@binary.path]
     command.unshift(@binary.xvfb_run_path) if options[:use_xvfb]
-    command += parse_options(options)
+    command += option_parser.parse(options)
     command << url
     command << generated_pdf_file.path.to_s
 
@@ -61,9 +62,8 @@ class WickedPdf
     if track_progress?(options)
       invoke_with_progress(command, options)
     else
-      err = Open3.popen3(*command) do |_stdin, _stdout, stderr|
-        stderr.read
-      end
+      _out, err, status = Open3.capture3(*command)
+      err = [status.to_s, err].join("\n") if !err.empty? || !status.success?
     end
     if options[:return_file]
       return_file = options.delete(:return_file)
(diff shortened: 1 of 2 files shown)


andrebnf (approved):
Fixed bug in production environment as listed here: https://github.com/mileszs/wicked_pdf/issues/1031#issuecomment-1287080936

Edit @ 2023-01-08 - forked version working fine since the original comment date

iggant:
hi, we really need this PR in the gem. Can we please merge it and release new version?

lib/wicked_pdf.rb
```diff
@@ -61,9 +62,8 @@ def pdf_from_url(url, options = {})
     if track_progress?(options)
       invoke_with_progress(command, options)
     else
-      err = Open3.popen3(*command) do |_stdin, _stdout, stderr|
-        stderr.read
-      end
+      _out, err, status = Open3.capture3(*command)
+      err = [status.to_s, err].join("\n") if !err.empty? || !status.success?
```
unixmonkey, line 66: Why the change from `popen3` to `capture3` here, and the joining?

lib/wicked_pdf.rb
```diff
@@ -79,6 +79,7 @@ def pdf_from_url(url, options = {})
   rescue StandardError => e
     raise "Failed to execute:\n#{command}\nError: #{e}"
   ensure
+    clean_temp_files
```
unixmonkey, line 82: 👍

lib/wicked_pdf/option_parser.rb
```diff
@@ -61,7 +61,7 @@ def parse_header_footer(options)
           r += make_options(opt_hf, [:line], hf.to_s, :boolean)
           if options[hf] && options[hf][:content]
             @hf_tempfiles = [] unless defined?(@hf_tempfiles)
-            @hf_tempfiles.push(tf = WickedPdf::Tempfile.new("wicked_#{hf}_pdf.html"))
+            @hf_tempfiles.push(tf = File.new(Dir::Tmpname.create(["wicked_#{hf}_pdf", '.html']) {}, 'w'))
```
unixmonkey, line 64: Is there a reason `WickedPdf::Tempfile` doesn't work here? I remember adding that because an earlier version of Ruby didn't allow tempfiles to be created with an extension, which I know is no longer the case, but don't remember when that changed.

unixmonkey (commented):
This is great! Sorry it took a bit for me to take a closer look.

I did post a few questions if you don't mind answering.

lib/wicked_pdf/option_parser.rb
thanhcuong1990: @unixmonkey `WickedPdf::Tempfile` still can created the temp file. But the files is deleted before the pdf file generated.

iggant:
@unixmonkey please, can you review and merge this PR?

unixmonkey:
This is now in version 2.7.0 of wicked_pdf. Thank you for your help!

PR #1046, opened 2023-01-24
CI: add Ruby 3.2 to the test matrix

orien:
Ruby 3.2 has [been released](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/)! Let's add it to the build matrix.

The Rubocop build step was failing, so I got the latest version running and regenerated the TODO configuration.


unixmonkey:
Thank you!

PR #1051, opened 2023-02-23
Bump rubocop to 1.46

drdee:
Hi!

Bumping rubocop to the most recent version as we are using wicked_pdf in production but it's preventing us from running a more recent version of rubocop that supports Ruby 3.2

### DIFF
--- a/wicked_pdf.gemspec
+++ b/wicked_pdf.gemspec
@@ -34,7 +34,7 @@ DESC
   spec.add_development_dependency 'mocha', '= 1.3'
   spec.add_development_dependency 'rails'
   spec.add_development_dependency 'rake'
-  spec.add_development_dependency 'rubocop', '~> 1.24'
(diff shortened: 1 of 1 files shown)


unixmonkey:
Happy for the update, but curious. Since it's just a development dependency, how it is preventing you from upgrading?

drdee:
In our Gemfile.lock, rubocop gets pinned to the version specified by wicked_pdf and the `~>` creates a very narrow range of acceptable rubocop versions.

unixmonkey:
Thanks. I'll have to check that out.

drdee:
Thanks! Forgot, should we bump the version?

unixmonkey:
You don't have to, but if you'd like to help me get a gem release out, I would appreciate if you could add merged, but unreleased PRs to the CHANGELOG, and bump the version in a separate PR.

PR #1058, opened 2023-04-25
Update pdf_helper.rb to use lookup_context to fetch template

### LINKED ISSUE #1057: Missing usage of view_lookup to fetch template
Currently the template for render view by controller is made by fetching directly the virtual path like `controller/action`
https://github.com/mileszs/wicked_pdf/blob/b116c432c9aab33cae58bc8d2045b240d7c7ec78/lib/wicked_pdf/pdf_helper.rb#L81

This can be an issue when we are in a structure where there is a couple of nesting controllers like:
- `ApplicationController`
- `Admin::BaseController < ApplicationController`
- `Admin::OrdersController < Admin::BaseController`

With a views folder like 
- ...
- `app/views/admin/base/show.pdf.erb`
- `app/views/admin/base/show.html.erb`

The controller use the `lookup_context` to fetch the correct `show` view thanks to the `self.local_prefixes`.
Let's for example try to render the show action in html format. The lookup context try to load the first available view traversing:
- `app/views/admin/orders`
- `app/views/admin/base`
- `app/views/application`

This should be the same as for the pdf format request but it seems not. To solve this we can directly use `lookup_context`

retsef:
Fix #1057

### DIFF
--- a/lib/wicked_pdf/pdf_helper.rb
+++ b/lib/wicked_pdf/pdf_helper.rb
@@ -78,7 +78,7 @@ class WickedPdf
     def make_and_send_pdf(pdf_name, options = {})
       options[:wkhtmltopdf] ||= nil
       options[:layout] ||= false
-      options[:template] ||= File.join(controller_path, action_name)
+      options[:template] ||= lookup_context.find(action_name, _prefixes, false).virtual_path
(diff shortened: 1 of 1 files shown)


Joyheaven7 (approved):
Ggh

PR #1061, opened 2023-05-02
require stringio before use

wizardofelves:
in non rails environments we need to require stringio before its use.


unixmonkey:
Why was this closed?

Issue #1066, opened 2023-07-04
Gem breaks with Shakapacker v 7.0.


sdrioux:
## Issue description

This gem does not work with the latest version of Shakapacker.  In version 7.0 all usages of "webpacker" was changed to "shakapacker".  As a result, this line results in an error: `cannot load such file -- webpacker/version (LoadError)`.  There may be other name issues hiding behind that one, as this gem refer to the `Webpacker` module in other places, however in this case the version.rb file was moved to `shakapacker/version.rb`

## System specifications

wicked_pdf gem version (output of `cat Gemfile.lock | grep wicked_pdf`): 2.6.3

Webpacker version 7.0.1

unixmonkey:
This is now in version 2.7.0 of wicked_pdf. Thank you for your help!

PR #1067, opened 2023-07-13
Shakapacker 7 compatibility (#1066)

Gregurevic:
Hi!
### Issue
I'm using `wicked_pdf` in a project where assets are being delivered with `shakapacker` version `7` .
I ran into an issue where the following error message was raised when I tried to use pack asset helpers:
```
cannot load such file -- webpacker/version (LoadError)
```
I've seen that this issue has already been reported here: https://github.com/mileszs/wicked_pdf/issues/1066
### Cause
I've found out that this line was raising the error:
`lib/wicked_pdf/wicked_pdf_helper/assets.rb :237`
```ruby
require 'webpacker/version'
```
Shakapacker started renaming their entire project from `webpacker` to `shakapacker` .
See the changelog for details: https://github.com/shakacode/shakapacker/blob/master/CHANGELOG.md#changed
> _Rename Webpacker to Shakapacker in the entire project including config files, binstubs, environment variables, etc. with a high degree of backward compatibility._
### Fix
A possible fix could have been to just remove the line raising the error. I believe that when `defined?(Webpacker)` is true, we do not need to `require 'webpacker/version'`. It will already be defined.

I figured tho that a more long lasting solution would be to check if either of Shakapcker or Webpacker is defined. In case shakapacker's backward compatibility expires.
### Tests
I did not include any tests in my PR. The current test suite doesn't have the ability to test assets delivered by shakapacker. I believe that the CI test matrix would need to be expanded and the gemfiles would need to be reworked to do so. I've found that to be a lot larger of a fix than what I originally intended to do.

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -231,12 +231,11 @@ class WickedPdf
       end
 
       def webpacker_version
-        return unless defined?(Webpacker)
-
-        # If webpacker is used, need to check for version
-        require 'webpacker/version'
-
-        Webpacker::VERSION
+        if defined?(Shakapacker)
+          Shakapacker::VERSION
+        elsif defined?(Webpacker)
+          Webpacker::VERSION
+        end
       end
     end
   end


lib/wicked_pdf/wicked_pdf_helper/assets.rb
```diff
@@ -231,12 +231,11 @@ def running_in_development?
...
-        return unless defined?(Webpacker)
-
-        # If webpacker is used, need to check for version
-        require 'webpacker/version'
-
-        Webpacker::VERSION
+        if defined?(Shakapacker)
+          Shakapacker::VERSION
+        elsif defined?(Webpacker)
+          Webpacker::VERSION
```
sdrioux, line 237: Did you run your app locally with this fix? Aren’t there other places where the Webpacker constant is being referenced?  Curious if this would fix the initial error but make way for another.

Gregurevic: Hi!

Yes I monkey patched it in my app, and it is working properly, running in production.

As I have stated in my PR description above:
- The issue is caused by this line: `require 'webpacker/version'`
- The issue could be resolved by just removing this line. (`webpacker/version` would be present anyway, no need to require it again.)
- To clarify once more, you will not get an error when referencing `Webpacker` in code, but you will get one when you try to require `webpacker/version` .

This is happening because Shakapacker decided to slowly rename everything "webpacker" to "shakapacker" in their codebase. At the moment they are trying to provide backwards compatibility.
- This backwards compatibility is the reason why referencing `Webpacker` doesn't raise an error. When using `shakapacker` version `7` .
- But since they already renamed much of their source files, you can not directly require `webpacker/version`, because no such file exists. You would need to require `shakapacker/version` under `shakapacker` version `7` .
  - This is present in the project: https://github.com/shakacode/shakapacker/blob/v7.0.2/lib/shakapacker/version.rb
  - This is not: https://github.com/shakacode/shakapacker/blob/v6.6.0/lib/webpacker/version.rb

If you do not like the idea of removing that require statement, then I can change my fix to the following:
```ruby
def webpacker_version
  if defined?(Shakapacker)
    require 'shakapacker/version'          # add new require statement for shakapacker version 7+
    Shakapacker::VERSION
  elsif defined?(Webpacker)
    require 'webpacker/version'            # keep the old require statement
    Webpacker::VERSION
  end
end
```

To clarify once more, wicked_pdf aside from this one line of code ( `require 'webpacker/version'` ) is `shakapacker` version `7` compatible. If we remove or change this line in the above mentioned way, then we end up with full compatibility.

This full compatibility is the goal of my current PR.

sdrioux:
Oh I see, I didn't realiize the `Webpacker` constant was still supported.  Great!  How do we get this released?

sdrioux (approved):

unixmonkey:
@sdrioux Can you confirm everything still works as expected if you install this branch into your project like:
__Gemfile__
```ruby
gem 'wicked_pdf', github: 'Gregurevic/wicked_pdf', ref: '995126b441db21eb39ff964fc3fee9e2babe08ed'
```

AJFaraday:
@unixmonkey I've been looking at this same issue and I can confirm that by using that line in my gemfile it worked as expected.

unixmonkey:
Thank you for confirming this @AJFaraday .

lib/wicked_pdf/wicked_pdf_helper/assets.rb
sk-: @unixmonkey @AJFaraday I'm not sure how this was tested, but this is not working for us. We currently have version 2.6.3, and the change to 2.7.0 fails with an uninitialized constant Webpacker::VERSION error.

The same happens when I load the rails console and try to access the constant.

Note that nowhere in the webpacker code I see the require of the version file.

unixmonkey: @sk- I would be happy to take a fix PR.

PR #1068, opened 2023-08-18
feat: Add keep_temp flag

crespire:
Adds an option to keep the temporary files generated by `pdf_from_string` method. It defaults to using the `close!` handler in the ensure block.

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -43,7 +43,8 @@ class WickedPdf
     string_file.write_in_chunks(string)
     pdf_from_html_file(string_file.path, options)
   ensure
-    string_file.close if string_file
+    close_method = options.key?(:keep_temp) ? :close : :close!
+    string_file.send(close_method)
   end
 
   def pdf_from_url(url, options = {}) # rubocop:disable Metrics/CyclomaticComplexity


crespire:
For compatibility purposes, we might consider switching the option to `delete_temp` and adjusting the logic to match as that's the current default behaviour. Happy to do so, let me know!

unixmonkey:
Yeah, I think I'd prefer the option to be `delete_temporary_files: true`, and check the value rather than key presence, & use a conditional rather than `send` something more like this:
```ruby
if options[:delete_temporary_files]
  file.close!
else
  file.close
end
```

What do you think?

crespire:
> Yeah, I think I'd prefer the option to be `delete_temporary_files: true`, and check the value rather than key presence (someone might set it to something that evaluates to `false`), & use a conditional rather than `send` something more like this:
> 
> ```ruby
> if options[:delete_temporary_files]
>   file.close!
> else
>   file.close
> end
> ```
> 
> What do you think?

I think that's very sensible! I'll make the changes.

unixmonkey:
Could you please add this new option to the README's list of "all options" please?

crespire:
As an entry in this section? https://github.com/mileszs/wicked_pdf#advanced-usage-with-all-available-options

unixmonkey:
Great stuff! Thank you.

unixmonkey:
This is now in version 2.7.0 of wicked_pdf. Thank you for your help!

jjb:
Thanks for what seems to be a nice feature

maybe there could be more docs on when/why this is useful

without this, the temp files will be cleaned up during GC, right?

so it's useful maybe for someone creating pdfs at high volume, where the disk files might add up quickly?

PR #1074, opened 2023-09-05
fix: check if file exists before deleting

### LINKED ISSUE #1073: No such file or directory @ apply2files
## Issue description

After upgrading to 2.7.0 version, the pdf generation fails on non-dev/test environment with the following error:

```
No such file or directory @ apply2files - /tmp/wicked_header_pdf20230905-21-mm5ta8.html
```

backtrace:

```
vendor/bundle/ruby/2.7.0/gems/wicked_pdf-2.7.0/lib/wicked_pdf.rb:112 delete
vendor/bundle/ruby/2.7.0/gems/wicked_pdf-2.7.0/lib/wicked_pdf.rb:112 block in clean_temp_files
vendor/bundle/ruby/2.7.0/gems/wicked_pdf-2.7.0/lib/wicked_pdf.rb:112 each
vendor/bundle/ruby/2.7.0/gems/wicked_pdf-2.7.0/lib/wicked_pdf.rb:112 clean_temp_files
vendor/bundle/ruby/2.7.0/gems/wicked_pdf-2.7.0/lib/wicked_pdf.rb:85 pdf_from_url
vendor/bundle/ruby/2.7.0/gems/wicked_pdf-2.7.0/lib/wicked_pdf.rb:36 pdf_from_html_file
vendor/bundle/ruby/2.7.0/gems/wicked_pdf-2.7.0/lib/wicked_pdf.rb:44 pdf_from_string
```

The issue seems to be related with the `#clean_temp_files` method.

## Expected or desired behavior

- pdf generation succeeds
- `#clean_temp_files` checks if file exists before trying to delete

## System specifications

wicked_pdf gem version (output of `cat Gemfile.lock | grep wicked_pdf`): 2.7.0

wkhtmltopdf version (output of `wkhtmltopdf --version`): 0.12.6.1 (with patched qt)

whtmltopdf [provider gem](https://rubygems.org/search?utf8=%E2%9C%93&query=wkhtmltopdf) and version if one is used:

platform/distribution and version (e.g. Windows 10 / Ubuntu 16.04 / Heroku cedar): Alpine

LuanGB:
fixes #1073

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -109,6 +109,9 @@ class WickedPdf
   def clean_temp_files
     return unless option_parser.hf_tempfiles.present?
 
-    option_parser.hf_tempfiles.each { |file| File.delete(file) }
+    option_parser.hf_tempfiles.each do |file|
+      next unless File.exists?(file.path)
+      File.delete(file)
+    end
   end
 end
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -110,7 +110,8 @@ class WickedPdf
     return unless option_parser.hf_tempfiles.present?
 
     option_parser.hf_tempfiles.each do |file|
-      next unless File.exists?(file.path)
+      next unless File.exist?(file.path)
+
       File.delete(file)
     end
   end


unixmonkey:
FYI, this issue also may exist here: https://github.com/mileszs/wicked_pdf/blob/master/lib/wicked_pdf.rb#L46-L50 where the files are closed without checking they exist first.

I'm curious how this is a new problem after the upgrade though... It seems that if any other process clearing tempfiles (a deploy maybe?) could cause this to happen in both old and new versions.

It might also be nice to delete or preserve the header & footer tempfiles with that same `:delete_temporary_files` option and call `close` or `close!` based on it.

LuanGB:
I'm suspecting that this may be related to this PR: https://github.com/mileszs/wicked_pdf/pull/1039
Cannot put my finger on what exactly is causing it, however

unixmonkey:
BTW, although it is very tiny, there could be a race-condition where the file exists while checking if the file exists and deleting it. Maybe it'd be better if closing a tempfile were `rescue`-able than checking for existence. Should speed things up too.

LuanGB:
That's a good idea. I'll put something together soon.

unixmonkey:
Ping.

alex-kovshovik:
I ran into the same issue, probably will have to use the forked version until this fix is released.

PR #1093, opened 2024-01-05
Improve running tests locally

d4rky-pl:
This PR improves local developer experience by pointing out what needs to be done for the tests to work locally without relying on CI

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,6 +23,13 @@ end
 
 desc 'Setup and run all tests'
 task :setup_and_run_tests do
+  ENV['BUNDLE_GEMFILE'] = File.join(__dir__, 'gemfiles', "#{ENV['RAILS_VERSION']}.gemfile") if ENV['RAILS_VERSION']
+
+  unless ENV['BUNDLE_GEMFILE']
+    puts 'You need to specify either BUNDLE_GEMFILE pointing to a Gemfile in gemfiles/ directory or RAILS_VERSION or the tests will fail.'
+    exit 1
+  end
+
   Rake::Task[:dummy_generate].invoke unless File.exist?('test/dummy/config/environment.rb')
   Rake::Task[:test].invoke
 end
--- a/Rakefile
+++ b/Rakefile
@@ -23,12 +23,14 @@ end
 
 desc 'Setup and run all tests'
 task :setup_and_run_tests do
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -23,6 +23,13 @@ end
 
 desc 'Setup and run all tests'
 task :setup_and_run_tests do
+  ENV['BUNDLE_GEMFILE'] = File.join(__dir__, 'gemfiles', "#{ENV['RAILS_VERSION']}.gemfile") if ENV['RAILS_VERSION']
```
mathieujobin, line 26: maybe it could default to the latest available rails Gemfile and display that it has done that automatically instead of exiting and forcing the user to set an ENV var ?

d4rky-pl: @mathieujobin I've switched the Rakefile and GH actions to use `RAILS_VERSION` directly as otherwise `BUNDLE_GEMFILE` was defaulting to `wicked_pdf/Gemfile` which never actually works

Rakefile
```diff
@@ -23,13 +23,15 @@ end
...
 
-  unless ENV['BUNDLE_GEMFILE']
-    puts 'You need to specify either BUNDLE_GEMFILE pointing to a Gemfile in gemfiles/ directory or RAILS_VERSION or the tests will fail.'
-    exit 1
+  unless ENV['RAILS_VERSION']
+    rails_version = Dir.glob(File.join(__dir__, 'gemfiles', '*.gemfile')).max[%r{gemfiles/(.*?)\.gemfile}, 1]
+    puts "RAILS_VERSION has not been specified, defaulting to the latest available Rails version (#{rails_version})"
   end
 
+  ENV['BUNDLE_GEMFILE'] = File.join(__dir__, 'gemfiles', "#{rails_version}.gemfile")
```
mathieujobin, line 33: interesting

mathieujobin: I have switch multiple projects to remove all of the individual gemfiles and use ENV var inside the main gemfiles instead... for some reason, it fails for this project. but I have this WIP branch if you want to take a look 

https://github.com/mileszs/wicked_pdf/compare/master...mathieujobin:wicked_pdf:ruby33

n-rodriguez:
Hi there! Why not using https://github.com/thoughtbot/appraisal ? It's the main purpose of this gem :wink:

PR #1096, opened 2024-01-13
Fix Propshaft encoding issue

yenshirak:
We get the following error after replacing Sprockets with Propshaft:
`ActionView::Template::Error (incompatible character encodings: ASCII-8BIT and UTF-8)`

This fix resolves the issue.

Propshaft::Asset#content returns an ASCII-8BIT encoded string because it's using File::binread under the hood so the encoding needs to be set to UTF-8 if it's a JS or CSS file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [master branch] - Unreleased
 ### Breaking Changes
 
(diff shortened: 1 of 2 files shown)


yenshirak:
@mathieujobin done

mathieujobin:
@unixmonkey @mileszs Can we cut a release with this fix ?

unixmonkey:

Issue #1098, opened 2024-01-24
version 2.7.0 breaks with webpacker 


rioug:
## Issue description
When trying to generate a PDF with `wicked_pdf` 2.7.0 it breaks with the following error : 
 
```
 ActionView::Template::Error:
       uninitialized constant Webpacker::VERSION
     
                 Webpacker::VERSION
```
The issue was introduced in this PR https://github.com/mileszs/wicked_pdf/pull/1067

## Expected or desired behavior
It should generate a pdf without any error :)

## System specifications

wicked_pdf gem version (output of `cat Gemfile.lock | grep wicked_pdf`): 2.7.0

webpacker version: 5.4.4 ( I don't think it has anything to do with the webpacker version but who knows ? )

wkhtmltopdf version (output of `wkhtmltopdf --version`): wkhtmltopdf 0.12.6 (with patched qt)

whtmltopdf [provider gem](https://rubygems.org/search?utf8=%E2%9C%93&query=wkhtmltopdf) and version if one is used: wkhtmltopdf-binary (0.12.6.6)

platform/distribution and version (e.g. Windows 10 / Ubuntu 16.04 / Heroku cedar):  Ubuntu 22.04.3

PR #1099, opened 2024-01-24
Fix webpacker compatibility

rioug:
As explain in #1098 this PR https://github.com/mileszs/wicked_pdf/pull/1067 broke compatibility with `Webpacker`

## Cause 
The following line has been removed, resulting in an error when trying to access `Webpacker::VERSION`
```ruby
require 'webpacker/version'
```
## Fix
Re add the `require 'webpacker/version'` and for consistency also add `require 'shakapacker/version'` when using `Shakapacker`

## Test
I tested it on our code base and our test suite is green. 
Ideally someone using `Shakapacker` would give this ago as well

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -232,8 +232,10 @@ class WickedPdf
 
       def webpacker_version
         if defined?(Shakapacker)
+          require 'shakapacker/version'
           Shakapacker::VERSION
         elsif defined?(Webpacker)
+          require 'webpacker/version'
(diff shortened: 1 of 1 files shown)


mathieujobin (approved):

unixmonkey (approved):

rioug:
@unixmonkey Thanks for approving the PR !

I can see some the CI test failed but it doesn't seem to be related to the PR, should they be fixed ? 
Also I can get the test to run on my local environment but I can see CI is passing with Ruby 3.0 and Rails 7, so I won't be spending more time looking at it.

unixmonkey:

PR #1090, opened 2024-01-05
Add WickedPdf.configure method

kevinnio:
Adds the `WickedPdf.configure` method described in #1089. Also adds a deprecation warning for `WickedPdf.config=` which can be silenced.

PR #1105, opened 2024-02-27
Correct syntax for configure block in README.md

pcothenet:
The syntax for the `configure` block in the doc was incorrect.

As far as I can tell `WickedPdf` doesn't support a `configure` block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,10 @@ You can see what flags are supported for the current version in [wkhtmltopdf's a
 If your wkhtmltopdf executable is not on your webserver's path, you can configure it in an initializer:
 
 ```ruby
-WickedPdf.configure do |c|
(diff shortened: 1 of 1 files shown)


unixmonkey:
The syntax was changed pretty recently in https://github.com/mileszs/wicked_pdf/pull/1090 but not packaged up into a new gem version yet, mostly because I haven't written up all the changes in the CHANGELOG yet.

Thanks for reminding me this needs done. I'll get a new release out very soon.

pcothenet:
@unixmonkey that makes sense! Feel free to close this once done.

unixmonkey:
I've just published version 2.8.0 of the Wicked PDF gem, that includes the new `configure` block syntax.

Thank you for caring enough to contribute!

Please let me know if you have any issues.

PR #1106, opened 2024-02-27
Update Config Example in README and template

manewitz:
Hello! Just some minor tweaks so the config example in the README is valid syntax and options in the template can be uncommented without modification. Cheers!


unixmonkey:
Oh shoot! I missed those trailing commas and such. Thanks so much!

PR #1110, opened 2024-04-18
fix: OpenStruct causes NameError in Rake13.2.0

### LINKED ISSUE #1109: OpenStruct causes NameError in Rake13.2.0
https://github.com/mileszs/wicked_pdf/blob/345b12baa167d245124b876a241007896226b3e7/lib/wicked_pdf.rb#L34

I bump the version of Rake to 13.2.0 and got a NameError at the above point.

Rake uses Struct instead of OpenStruct in version 13.2.0.
https://github.com/ruby/rake/pull/545

It seems that OpenStruct loading is implicitly dependent on Rake.

I suggest adding `require 'ostruct'` or using Struct instead of the deprecated OpenStruct.

rhiroe:
resolved #1109

Loading of "ostruct" depends on Rake, so we need to explicitly load it and resolve the dependency.
Because Rake does not load "ostruct" since version 13.2.0.
https://github.com/ruby/rake/pull/545

### DIFF
--- a/lib/wicked_pdf.rb
+++ b/lib/wicked_pdf.rb
@@ -5,6 +5,7 @@ require 'logger'
 require 'digest/md5'
 require 'rbconfig'
 require 'open3'
+require 'ostruct'
 
 require 'active_support/core_ext/module/attribute_accessors'
 require 'active_support/core_ext/object/blank'


heliocola:
👋 Maybe this is worthy of a version 2.8.1.

I run into this issue a few weeks ago upgrading a Rails app from 6.1 to 7.0.8.1 and I ended up adding the `require 'ostruct'` to the application.rb.

This app was already using WickedPdf v2.8.0.

And interesting enough: the OpenStruct name error was only happening in Production (not staging, nor CI)... :-D

PR #1115, opened 2024-05-04
Fix: Ensure assets without extensions are handled correctly

d4rky-pl:
This PR fixes the issues reported after #1094 and related to the asset manifest implementation not looking for the assets properly.

---
As a side note, this will most likely be my last contribution to this project. The fix itself took me 5 minutes to implement but adding those four simple test cases took... 1.5h. I wish I was joking 🥲

If anyone still has energy it would be great to at least figure out how to upgrade Mocha to the newer version as this one has a bug that prevents stubs from clearing properly (though it would be amazing to get rid of it completely and switch to rspec)

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -216,7 +216,13 @@ class WickedPdf
         assets = Rails.application.assets_manifest.assets
 
         if File.extname(path).empty?
-          assets.find { |asset, _v| File.basename(asset, File.extname(asset)) == path }&.last
+          assets.find do |asset, _v|
+            directory = File.dirname(asset)
+            asset_path = File.basename(asset, File.extname(asset))
+            asset_path = File.join(directory, asset_path) if directory != '.'
+
+            asset_path == path
+          end&.last
         else
           assets[path]
         end
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -202,10 +202,23 @@ class WickedPdf
         elsif defined?(Propshaft::Assembly) && Rails.application.assets.is_a?(Propshaft::Assembly)
           PropshaftAsset.new(Rails.application.assets.load_path.find(path))
         elsif Rails.application.respond_to?(:assets_manifest)
-          asset_path = File.join(Rails.application.assets_manifest.dir, Rails.application.assets_manifest.assets[path])
+          relative_asset_path = get_asset_path_from_manifest(path)
+          return unless relative_asset_path
+
+          asset_path = File.join(Rails.application.assets_manifest.dir, relative_asset_path)
           LocalAsset.new(asset_path) if File.file?(asset_path)
         else
           SprocketsEnvironment.find_asset(path, :base_path => Rails.application.root.to_s)
+        end
+      end
+
+      def get_asset_path_from_manifest(path)
+        assets = Rails.application.assets_manifest.assets
+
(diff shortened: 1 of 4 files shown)


d4rky-pl:
@doits @joshuay03 would you like to test this PR in your apps and confirm this solved the issue for you?
@unixmonkey @mathieujobin let me know if you need anything else to land this patch once we get a confirmation it works

joshuay03:
> @doits @joshuay03 would you like to test this PR in your apps and confirm this solved the issue for you?

@d4rky-pl I can confirm that this fixes the issue we were having. Thank you for working on this, and the original PR, which seems like a super useful feature!

d4rky-pl:
@unixmonkey do you need anything else or can we release this as 2.8.1? :)

mathieujobin (commented):
this is a cool fix, and I can throw a CI run with my app running with your branch. but if I only have assets with explicit extensions, it won't change a thing, if my code review is correct.

so I wonder, what if we want extensions to be explicit? I don't know.

d4rky-pl:
@mathieujobin I have no idea either but requiring extensions would be an API breaking change and I wanted to avoid that (especially considering the library is more or less on life support at this point).

My solution is based partially on how Sprockets solve that (in `find_asset` -> `resolve` -> `resolve_logical_path` -> `resolve_under_paths`) except I didn't add prioritizing which asset to pick based on the mime type because we don't know it and can't assume when using `wicked_pdf_asset_base64` (as it's decoupled from the `stylesheet` / `javascript` concept).

d4rky-pl:
One other thing we could do is make it fail when there's more than one asset without extension to avoid unexpected situations (like having `fonts.css` and `fonts.js` and then `wicked_pdf_asset_base64` picking one but not the other) but that's also going to be a breaking change 🤔

alessandrostein:
Any plans to release `2.8.1`?

mathieujobin (approved):
my app tests pass with these changes. I don't have write access, otherwise I would merge and release. 👍🏽

lib/wicked_pdf/wicked_pdf_helper/assets.rb
```diff
@@ -202,13 +202,26 @@ def find_asset(path)
...
         else
           SprocketsEnvironment.find_asset(path, :base_path => Rails.application.root.to_s)
         end
       end
 
+      def get_asset_path_from_manifest(path)
+        assets = Rails.application.assets_manifest.assets
+
+        if File.extname(path).empty?
+          assets.find { |asset, _v| File.basename(asset, File.extname(asset)) == path }&.last
```
chaadow, line 219: Unfortunately this does not work for nested directories of assets.

for example `company/logo/some_variant.svg`
Here is the the fix:
```ruby
assets.find do |asset, _v| 
  directory = File.dirname(asset)
  filename_without_extension = File.basename(asset, File.extname(asset))

 (directory == "." ? filename_without_extension : File.join(directory, filename_without_extension)) == path
end&.last
```

d4rky-pl:
Added more tests and a fix, good catch @chaadow! Let me know if you can think of any other cases where this could break so I can add more tests

chaadow (approved):
Looks good to me. Thanks!

d4rky-pl:
@unixmonkey let me know if you need anything else to have this released. If there are any further bug reports related to this change, either before or after this PR is merged, feel free to @ me as well

chaadow:
@unixmonkey Hi! is it possible to release this fix 🙏🏼

unixmonkey (approved):
👍

unixmonkey:
This is now published in version `2.8.1`. Please let me know if you experience any issues after upgrading to this.

Thank you all for your help!

PR #1120, opened 2024-06-11
Allow/Fix precompiled assets with Propshaft pipeline

serggl:
## Purpose

[rails/propshaft](https://github.com/rails/propshaft) support was added with #1010
However it appears to only work with dynamic assets and fails with precompiled ones:

```
#<UncaughtThrowError: uncaught throw #<ActionView::Template::Error: undefined local variable or method `content' for an instance of WickedPdf::WickedPdfHelper::Assets::PropshaftAsset>>
```

## Solution

Improve the original solution and use Propshaft's built-in [resolver](https://github.com/rails/propshaft/blob/main/lib/propshaft/assembly.rb#L21) to handle gracefully both asset types

### DIFF
--- a/lib/wicked_pdf/wicked_pdf_helper/assets.rb
+++ b/lib/wicked_pdf/wicked_pdf_helper/assets.rb
@@ -28,20 +28,6 @@ class WickedPdf
         end
       end
 
-      class PropshaftAsset < SimpleDelegator
-        def content_type
-          super.to_s
-        end
-
-        def to_s
-          content
-        end
-
-        def filename
-          path.to_s
-        end
-      end
-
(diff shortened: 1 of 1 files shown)


serggl:
@unixmonkey please review 🙏🏽

thr3a:
I encountered the same error. I hope this gets merged.

```
undefined local variable or method `content' for nil:WickedPdf::WickedPdfHelper::Assets::PropshaftAsset
```

rafaeldev:
Is there any update about it?

mfittko:
@unixmonkey We also need this, could we please get this out?

unixmonkey:
Thanks!

PR #1123, opened 2024-07-11
Change "verified" versions in README

okuramasafumi:
https://github.com/mileszs/wicked_pdf/blob/master/.github/workflows/ci.yml
This CI matrix implies this gem works with newer Ruby/Rails versions.


unixmonkey:
Thank you!

